PR #29, opened 2016-07-26
Fixups and Markdown formatting

jlevy:
Seems like https://github.com/shurcooL/markdownfmt is the best option to keep things consistent. 

I also looked at https://github.com/slang800/tidy-markdown (which is nice since it has an Atom plugin) but it has quirks around nested itemized lists (some are condensed and some are loose), and it deliberately strips smart quotes, both of which seemed a bit questionable.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,15 +1,11 @@
-## Contributing
+Contributing
+------------
 
-Contributions of all kinds, including discussion, corrections, additions, and improvements, are welcome!
-We hope you'll join and help, in small ways or large.
-We gladly credit all contributors and authors.
-Here are few notes before you jump in.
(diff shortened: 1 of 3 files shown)


ThanosBaskous:
I'm on board with a formatter. @jlevy what do you think about not _requiring_ contributors to run it before every submission but instead, one of us occasionally running it to clean things up?

jlevy:
Yeah, agreed, good point. Obstacles to contributing should be minimal.

ThanosBaskous:
👍 Looks good to me, thanks @jlevy.

jlevy:
Linking the tidy-markdown lists issue, for reference.
https://github.com/slang800/tidy-markdown/issues/17

PR #38, opened 2016-08-11
Info on new Application Load Balancer support.

jlevy:


ThanosBaskous:
@jlevy 👍 great turnaround time.

This is a good starting place.

I can take on doing the following in a subsequent PR, if you are on board:

- Rename "ELBs" section to "Load Balancers"
- Segment "Load Balancers" section into 
    - Common load balancer basics / tips
    - ELB-specific notes and tips
    - ALB-specific notes and tips
- Add notes to ALB section (I've done some testing on it already (via beta access) and some more to add. I can expand on these here if you want or we can wait until the PR.
- (Nit) Introduce and use "ALB" abbreviation so that searchers (and search engines) will find it.

jlevy:
Awesome. Agree on all items. For simplicity merged but will let you take it on those with a new PR. Tips from your beta experience would be great.

PR #85, opened 2016-10-08
Subsection links and cleanup

### LINKED ISSUE #76: Better TOC and subsection links
Some sort of better solution (bulleted list or table) to make it easier and more clickable to jump down the doc from the first page.

ThanosBaskous:
- Convert main service block in TOC to table with links to most common subsections.
- Break AMI tips into basics/tips/gotchas
- Move some Autoscaling into tips into gotchas
- Add CLB section, basics subsection
- Add ALB section
- Add CLB and ALB to table of contents table (after load balancers)
- Fix "VPC and Netork Security Gotchas and Limitations" subheader typo

Resolves #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,29 +15,35 @@ The Open Guide to Amazon Web Services
 	-	[Managing AWS](#managing-aws)
 	-	[Managing Servers and Applications](#managing-servers-and-applications)
 -	*Specific AWS Products*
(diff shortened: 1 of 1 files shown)


jlevy:
👍 Looks great.

I wonder if we should consider making the whole TOC one table... but we can experiment with that later.

PR #97, opened 2016-10-11
document IPv4 & IPv6 access to S3

tomschlick:


jlevy:
Good one! LGTM. @max-zanko review?

tomschlick:
added the link

PR #101, opened 2016-10-11
Replace recommendation of GUI application for S3. Transmit is outdate…

dkocher:
…d,  missing AWS4 signature version support (crucial for new regions)  and multipart uploads besides ACLs, versioning, lifecycle configuration, storage classes and server side encryption.


paultyng:
Probably worth disclosing you work for the company that writes Cyberduck?

dkocher:
@paultyng Sure, you can see from my GitHub profile.

scheesman:
Transmit works fine for S3 on MacOS, so I disagree with replacing the recommendation.  Why not just add a secondary suggestion instead of recommending the removal of a competitor's product because you've deemed it "outdated"?

dkocher:
@scheesman Sure, you can leave the recommendation for Transmit. But it will not even connect to regions requiring AWS4 signatures.

max-zanko:
@dkocher, let's keep Transmit with a disclaimer on the missing/broken functions. I agree with @scheesman that the tool may be just fine for the basic use cases.

jlevy:
Agreed. Tagging with "neutrality" as it's good to track when extra care is needed about changes.

jlevy:
👍 This is a good addition, and thanks @dkocher @paultyng @scheesman @max-zanko for the discussion and improvements.

PR #121, opened 2016-10-12
EFS is GA, not Preview

shott85:
Per: https://aws.amazon.com/about-aws/whats-new/2016/06/amazon-elastic-file-system-efs-is-now-generally-available/


jlevy:
Yeah, weird how their release notes are off. But it is indeed GA now. So thanks.

PR #124, opened 2016-10-12
Update the definition of AWS SWF

xethorn:
First, thanks for doing this!

Just a small update regarding SWF: SWF is not only a managed background job workflow (you have to write your own decider and your own tasks). It also stores and tracks states of each activity (defined has “events" in the workflow's "history”), which it sends to the decider and the workers. I added polyglot because at the comparison of other systems like Luigi, SWF is language agnostic. You can write the decider in one language, and write an activity in Go, and another one in python.


ThanosBaskous:
lgtm. Thanks for your contribution, @xethorn!
@jlevy would you mind checking this out for final review/merge?

jlevy:
👍 LGTM

jlevy:
@xethorn Do you have a lot of experience with SWF? Any interest in contributing more on it?

jlevy:
And thanks!

mortaliorchard:
@jlevy : I'll try to. I wrote in python [Garcon](http://github.com/xethorn/garcon) which is used in production by a few companies.

xethorn:
@jlevy : I'll try to. I wrote in python [Garcon](http://github.com/xethorn/garcon) which is used in production by a few companies.

PR #129, opened 2016-10-12
Remove extra word

maiki:
It said, "shift shift". That's 100% too many shifts!


ThanosBaskous:
👍 Thanks for the fix, @maiki!

PR #142, opened 2016-10-12
Fix internal link to Learning section

StevenMaude:
Thought my browser was broken. Turns out it was the link.


jlevy:
👍 Thanks for the catch!

PR #147, opened 2016-10-12
Added two limitations on Lambda

theist:
Added two limitations on lambda

### DIFF
--- a/README.md
+++ b/README.md
@@ -1038,6 +1038,9 @@ Lambda
 -	🔸Lambda is a new technology. As of mid 2016, only a few companies are using it for large-scale production applications.
 -	🔸Managing lots of Lambda functions is a workflow challenge, and tooling to manage Lambda deployments is still immature.
 -	🔸AWS’ official workflow around managing function [versioning and aliases](https://docs.aws.amazon.com/lambda/latest/dg/versioning-aliases.html) is painful.
+-	🔸Lambda Functions must end in less than five minutes. If it is still running after that time it will be destroyed, considered failed and relaunched up to three times.
+-	🔸Any pending threads existing at the end of the caller function gets killed and the lambda is considered ended. If you raise background process (manually in python or by the normal node.js flow) you must ensure all threads are joined before let the function end.
+
 
 🚧 [*Please help expand this incomplete section.*](CONTRIBUTING.md)
 


README.md
```diff
@@ -1038,6 +1038,9 @@ Lambda
 -	🔸Lambda is a new technology. As of mid 2016, only a few companies are using it for large-scale production applications.
 -	🔸Managing lots of Lambda functions is a workflow challenge, and tooling to manage Lambda deployments is still immature.
 -	🔸AWS’ official workflow around managing function [versioning and aliases](https://docs.aws.amazon.com/lambda/latest/dg/versioning-aliases.html) is painful.
+-	🔸Lambda Functions must end in less than five minutes. If it is still running after that time it will be destroyed, considered failed and relaunched up to three times.
+-	🔸Any pending threads existing at the end of the caller function gets killed and the lambda is considered ended. If you raise background process (manually in python or by the normal node.js flow) you must ensure all threads are joined before let the function end.
```
jlevy: Does this apply to threads and processes? Clarify that?

Link here too would be great.

Minor: python -> Python

README.md
```diff
@@ -1038,6 +1038,9 @@ Lambda
 -	🔸Lambda is a new technology. As of mid 2016, only a few companies are using it for large-scale production applications.
 -	🔸Managing lots of Lambda functions is a workflow challenge, and tooling to manage Lambda deployments is still immature.
 -	🔸AWS’ official workflow around managing function [versioning and aliases](https://docs.aws.amazon.com/lambda/latest/dg/versioning-aliases.html) is painful.
+-	🔸Lambda Functions must end in less than five minutes. If it is still running after that time it will be destroyed, considered failed and relaunched up to three times.
```
jlevy: Link documenting this limit?

ThanosBaskous: Nit: "functions" should not be capitalized - it isn't on the product page: https://aws.amazon.com/lambda/details/

README.md
ThanosBaskous: Grammar: "Any pending threads existing at the end of the caller function gets killed"  -> "Any pending threads existing at the end of the caller function get killed".

Phrasing suggestion: "If you raise background process (manually in python or by the normal node.js flow) you must ensure all threads are joined before let the function end." -> "If you run background processes (manually in python or using the normal node.js flow), you must ensure that all threads are joined before letting the function end."

theist: @jlevy I don't know if this is documented, it comes from my own experience with lambdas. Any work left on background terminates when the lambda handler function reach its end

jlevy:
Thanks for fixing all our nit-picks and keeping it polished. :)

happysheaven:
Welcome

README.md
theist: mmm I said that that happens on a single lambda execution I was completely unaware of the all the freeze/reuse stuff. And I can't imagine a use case that can exploit this behaviour that it is not guaranteed to work in subsequent Lambda calls.

Perhaps if I rephrase it as:

"When the handler function ends the Lambda invocation is considered finished and it exits leaving any work pending undone. It is not guaranteed that these works can be recovered. Is advisable to wait for any threads/processes before the end of the handler function"

Better?

jlevy: @theist That sounds good to me. @ehammond may have further thoughts.
Let's also include the link from @Java1Guy ? https://aws.amazon.com/blogs/compute/container-reuse-in-lambda/
We could do a bit more explanation of container reuse as a separate point, too. Adding #256 to track.

jlevy:
@theist would be awesome if you can update/prep this for merge. Thanks!

ehammond:
> Lambda Functions must end in less than five minutes.

Incomplete. Each Lambda function has a timeout length defined by the user. The current maximum timeout length is 5 minutes.

> If it is still running after that time it will be destroyed, 

This wording is not accurate. The function is not "destroyed".

> and relaunched up to three times.

This is only true for some of the ways that Lambda functions are invoked. For some, it does not retry at all. For others, the Lambda function will continue to be retried forever. http://docs.aws.amazon.com/lambda/latest/dg/retries-on-errors.html

> Any pending threads existing at the end of the caller function gets killed

I have my doubts this is true. It certainly is not true for processes.

> If you raise background process (manually in python or by the normal node.js flow) you must ensure all threads are joined before let the function end.

Not true. @jlevy 's link explains how the Lambda function environment works. It is not simple to explain or understand. https://aws.amazon.com/blogs/compute/container-reuse-in-lambda/

jlevy:
Sounds like we need to revisit/rewrite this. @theist feel free to update this PR if you want to handle it, or @ehammond you're welcome to submit a new PR if you want to tackle it. Sounds a bit complex/confusing, but that is a good reason for us to cover it. Thanks to you both!

jlevy:
Of course, discussing on Slack might also help.

theist:
There was a lot of controversy and all the tips were questioned, reworded or claimed to be not enough accurate. Now the PR conflicts mainstream. I'll be happy even if this PR is closed without merge, because here is a lot of good points to generate better PRs than mine. I Don't see the point to fix a PR which isn't good enough from its conception. So I'm gona close this. Thanks!

happysheaven:
Fiverr
    
    
            
      {
          "@context": "http://schema.org",
          "@type": "EmailMessage",
          "action": {
            "@type": "ViewAction",
            "url": "https://www.fiverr.com/s2/c2099c6182",
            "name": "Browse Services on Fiverr"
          },
          "description": "Browse Services on Fiverr"
      }
    

        
            
                
                    
                        
                            
                                
                                    
                                        
                                    
                                
                                
                                    
                                    
                                
                                
                                    

                                        
    

    
        Happysheaven is tired of keeping Fiverr a secret.
    

    
        Fiverr is the world’s largest marketplace for services,where you can get everything you need at an unbeatable value.
    

    
        Sounds too good to be true? See for yourself!
    



                                        
                                            
                                                
                                                    
                                                        
                                                            Browse Services On Fiverr
                                                        
                                                    
                                                
                                                
                                                    
                                                        

		
			
			
			The Fiverr Team
		
                                                    
                                                
                                            
                                        
                                    
                                
                            
                        
                    
                
                
                    
                        
                            
                                

                                    © Fiverr International Limited 2010-2016, Fiverr Inc. | 401 Broadway, Suite 1600 New York NY, 10013 USA.
                                    Your personal information will not be stored or used for any marketing purposes.
                                
                            
                         
                    
                
            
        
            

    

If you would like to unsubscribe and stop receiving these emails click here.

PR #111, opened 2016-10-11
Update Azure info

esell:
This PR cleans up some of the Azure information, possibly closing issue #104. It also removes the blurb about GCE being the next closest competitor to AWS as the data that is linked to in the report does not support that statement.

Disclaimer: I'm a Microsoft employee :)

PR #156, opened 2016-10-12
Adding CenturyLink services

jorn-k:
CLC  (CenturyLink Cloud; former Tier3 offering) often considered in top 5 by analysts and pundits; should be included for comparison.


jlevy:
Thanks for the suggestions.

Seems like CenturyLink is a significant provider, but I'm not sure it's at a level where it'd be mentioned between Azure and Rackspace. Do you have some market sector data you could share? The Gartner report in #111 may help with this.

Also, do you mind disclosing if you have an affiliation with CenturyLink? The cloud ecosystem is full of competitors so we're looking to have a process where we disclose/review this kind of thing thoughtfully.

rjhintz:
Centurylink fell out of the top right in Gartner's Magic Quadrant in 2016 and has widely been reported to be trying to exit their Saavis acquisiton. News today is that they're acquiring Level3, so more telecom, less XaaS as a portion of their biz. 

I don't strong feelings about including anyone, besides Google and Azure, but the rest seem niche players at best.

QuinnyPig:
Suggest closing this; this is the first place I've ever seen CenturyLink offering something that isn't bandwidth.

DavidTPate:
I agree that I think this shouldn't be included and I would prefer to just reference something such as Gartner's Magic Quadrant to keep this guide more agnostic. Ideally, we would remove the "Rackspace" portion as well or at the very least only mention those that are in the top right of Gartner (which I expect includes Rackspace among others).

QuinnyPig:
Closing for now; feel free to continue the conversation / revisit this as the landscape changes.

PR #159, opened 2016-10-12
AZ naming gotcha example

ap-hyperbole:
Adding example to explain AZ naming as it was confusing to understand :

> the letters are assigned to physical AZs randomly per account


jlevy:
Thanks. I'm not convinced the AZ example is needed. It's a much longer way to say the same thing, and also it could be confusing as that region adds zones.

Closing for now but open to discuss if you think of a smaller change or explain what's unclear?

PR #165, opened 2016-10-12
Information about IAM session keys requiring a session token parameter.

richadams:


richadams:
📜  icon added. Amended previous commit to keep history tidy.

jlevy:
👍 Great addition, and great new emoji. :)

PR #166, opened 2016-10-12
Add information about Docker adoption rates

kky:


ThanosBaskous:
Thank you for your submission, @kky, your desire to improve the guide is much appreciated. I don't think that this information is specific enough to AWS to warrant inclusion in the guide and so I am recommending that we close this PR (assigning to other project lead @jlevy for confirmation). Feel free to respond if you want to discuss further.

jlevy:
Thanks @kky ! Some general info about Docker adoption might be helpful, but tend to agree with @ThanosBaskous that this doesn't quite make the cut on the guide to AWS, as it's not directly AWS related, and also because subjectively it seems a little heavy on the content marketing for Datadog.

Also, we do ask contributors mention it when they file PRs for products they are affiliated with.

In fact, Datadog _is_ a pretty widely used product, and we link to it already, albeit without a lot of commentary. We'd definitely be open to any (balanced) expansion in that area, e.g. how it fits in relative to other related products. Feel free to join our Slack and chat about this.

PR #172, opened 2016-10-12
Added tip on pre-empting instance retirement by stopping and starting the instance.

spudstuff:


spudstuff:
Added some extra detail, cheers.

PR #185, opened 2016-10-13
Update authors + roles

jlevy:


jlevy:
@forwidur @lynnlangit @richadams @donnemartin @merrells @benkehoe 

This reflects roles we've discussed, or that you're actually already filling. Ping me if you have questions/concerns. A bit more on editorial process is now [here](https://github.com/open-guides/og-aws/blob/master/CONTRIBUTING.md#editorial-process). Thanks so much!

PR #191, opened 2016-10-13
Expand the PaaS section of the service matrix

chids:
- Providers: AppFog and OpenShift
  - Build your own: Cloud Foundry and Convox


ThanosBaskous:
LGTM, thanks @chids - since this doesn't have an official owner, bouncing to @jlevy for second stage review.

jlevy:
👍 LGTM. Thanks @chids !
PaaS overall is potentially a big list but I think these are fair additions if we want large ones and illustrative examples.

Adding neutrality tag so we can track items where we want to be fair wrt vendors.

PR #193, opened 2016-10-13
Added v1 of information on Kinesis

KrishanBhasin:


KrishanBhasin:
I've added some details on Kinesis, and linked to some blogs that go into the things I've touched on in more detail.

README.md
```diff
@@ -1273,7 +1275,33 @@ EMR
...
 
+Kinesis Streams
+---
+
+### Kinesis Streams Basics
+
+-	📒 [Homepage](https://aws.amazon.com/kinesis/streams/) ∙ [Developer guide](https://docs.aws.amazon.com/streams/latest/dev/introduction.html) ∙ [FAQ](https://aws.amazon.com/kinesis/streams/faqs/) ∙ [Pricing](https://aws.amazon.com/kinesis/streams/pricing/)
+-	**Kinesis Streams** (which used to be only called Kinesis, before Kinesis Firehose and Kinesis Analytics were launched) is a service that allows you to injest high-throughput data streams for immediate or delayed processing by other AWS services
+- Kinesis Streams' subcomponents are called [Shards](https://docs.aws.amazon.com/streams/latest/dev/key-concepts.html). Each shard provides 1MB/s of write capacity and 2MB/s of read capacity at a maximum of 5 reads per second. A stream can have its Shards programatically increased or decreased based on a variety of metrics
+- All records entered into a Kinesis Stream are kept in time-order, but obviously it is up to the developer to ensure they are entered into the Stream in the correct order.
```
forwidur: W: This sentence seems a little bit unclear to me. I'd write it like that: "Each record is assigned a unique [sequence number](http://docs.aws.amazon.com/streams/latest/dev/key-concepts.html#sequence-number) after you write it to the stream. The records are ordered by that sequence number."

README.md
```diff
@@ -1273,7 +1275,33 @@ EMR
...
 
+### Kinesis Streams Gotchas and Limitations
+-	💸❗**Kinesis Streams are not included in the free tier!** Make sure if you do any experimentation with it on a personal account, you shut down the stream, or you can run up unexpected costs (~$11 per shard-month)
+-  Kinesis Streams' shards each only permit 5 reads per second. If you are using ```n``` shards in a particular stream, and evenly distributing your data across all of them, you will end up with a total of 5 reads per second. This is because a consumer cannot know which shard will contain new data, and will therefore need to check every single one. This means that there is a hard limit on the number of consumers you can have per stream, for any given latency.
+	- If you wish to have 5 consumers all reading data from one Stream with 5 shards, with a maximum permitted latency of 0.5 seconds, you will need to either split your data across two streams, or reduce your latency requirements.
+	- Each consumer will need to poll each shard once every 0.5 seconds, meaning each Shard will need to be queried 10 times a second - a value in excess of the maximum.
```
forwidur: Roll this into the previous paragraph.

KrishanBhasin:
Comments taken on-board and acted upon 👍

forwidur:
Awesome! Thanks @KrishanBhasin. 

One last thing - there seem to be conflicts. Could you please resolve those?

Thank you!

KrishanBhasin:
@forwidur, Done!

forwidur:
Thank you, @KrishanBhasin ! You rock!

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
 
+Kinesis Streams
+---
+
+### Kinesis Streams Basics
+
+-	📒 [Homepage](https://aws.amazon.com/kinesis/streams/) ∙ [Developer guide](https://docs.aws.amazon.com/streams/latest/dev/introduction.html) ∙ [FAQ](https://aws.amazon.com/kinesis/streams/faqs/) ∙ [Pricing](https://aws.amazon.com/kinesis/streams/pricing/)
+-	**Kinesis Streams** (which used to be only called Kinesis, before Kinesis Firehose and Kinesis Analytics were launched) is a service that allows you to injest high-throughput data streams for immediate or delayed processing by other AWS services
+- Kinesis Streams' subcomponents are called [Shards](https://docs.aws.amazon.com/streams/latest/dev/key-concepts.html). Each shard provides 1MB/s of write capacity and 2MB/s of read capacity at a maximum of 5 reads per second. A stream can have its Shards programatically increased or decreased based on a variety of metrics
```
jlevy: Shard -> shard

jlevy: Shard -> shard (per AWS usage)

jlevy: Also boldface **shard** since it's a definition.

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+- All records entered into a Kinesis Stream are assigned a unique Sequence Number as they are captured. The records in a Stream are ordered by this number, so any time-ordering is preserved.
```
jlevy: Sequence Number -> sequence number (per AWS usage)
Boldface.
Optional: A link on this too?

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+
+### Kinesis Streams Alternatives and Lock-in
+
+-	⛓Kinesis is most closely compared to [Apache Kafka](https://kafka.apache.org/), an open-source data injestion solution. It is possible to set up a Kafka cluster hosted on [EC2 instances](#ec2) (or any other VPS), however you are responsible for managing and maintaining both Zookeeper and the Kafka brokers in a highly availible configuration. Confluent have a good blog post on their recommendations on doing this [here](http://www.confluent.io/blog/design-and-deployment-considerations-for-deploying-apache-kafka-on-aws/), which has links on the bottom to several other blogs they have written on the subject.
```
jlevy: This should probably be 🚪 (alternatives) not ⛓ (lock-in).

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+-   Kinesis uses very AWS-specific APIs and terms (e.g. Shards), so you should be aware of the potential future costs of migrating away from it, should you choose to use it.
```
jlevy: ⛓ applies here.

A little confused by "and terms (e.g. Shards)" part. Terminology doesn't lock you in. Maybe drop this part?

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+
+### Kinesis Streams Tips
+
+-	The [KCL](https://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-kcl.html) (Kinesis Client Library) is a very useful Java program (wrapped in a Multi-Language Daemon that makes it useable in Java, Node, Python, Ruby and .NET programs) that provides very simple interfaces for clients to use when consuming data from a Kinesis Stream. It provides the skeleton for 3 basic functions - ```intitialise```, ```process-records```, and ```shutdown```. As a developer, all you need to do is set up the config file to point at the correct Kinesis Stream, and fill out the provided functions in order to start consuming data from Kinesis.
```
jlevy: "Multi-Language Daemon" -> "multi-language interface called MultiLangDaemon"

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+	- The KCL uses a DynamoDB table to keep track of which records have been processed by the KCL. This ensures that all records are processed 'at least once'. It is up to the developer to ensure that the program can handle doubly-processed records.
```
jlevy: Unicode quotes: “at least once”

README.md
jlevy: It looks like these are not the exact names, right? https://github.com/awslabs/amazon-kinesis-client/blob/master/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangProtocol.java
(also there are typographic variations?).

How about just saying: "It provides the skeleton for initializing, processing records, and shutting down."

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+	- The KCL also uses DynamoDB to keep track of other KCL 'workers'. It automatically shares the available Kinesis Shards across all the workers as equally as possible.
+
+### Kinesis Streams Gotchas and Limitations
+-	💸❗**Kinesis Streams are not included in the free tier!** Make sure if you do any experimentation with it on a personal account, you shut down the stream, or you can run up unexpected costs (~$11 per shard-month)
+-  Kinesis Streams' shards each only permit 5 reads per second. If you are using ```n``` shards in a particular stream, and evenly distributing your data across all of them, you will end up with a total of 5 reads per second. This is because a consumer cannot know which shard will contain new data, and will therefore need to check every single one. This means that there is a hard limit on the number of consumers you can have per stream, for any given latency.
```
jlevy: Is there at a link on the limit of 5? AWS doc and/or another link?

Add 🔸.

Can you say this more tersely? "If you are evenly distributing your data across many shards, you may end up with only 5 reads per second in aggregate, if a consumer must check every shard to see which contains new data. This means..."

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+	- The KCL uses a DynamoDB table to keep track of which records have been processed by the KCL. This ensures that all records are processed 'at least once'. It is up to the developer to ensure that the program can handle doubly-processed records.
+	- The KCL also uses DynamoDB to keep track of other KCL 'workers'. It automatically shares the available Kinesis Shards across all the workers as equally as possible.
+
+### Kinesis Streams Gotchas and Limitations
+-	💸❗**Kinesis Streams are not included in the free tier!** Make sure if you do any experimentation with it on a personal account, you shut down the stream, or you can run up unexpected costs (~$11 per shard-month)
```
jlevy: Good note, but perhaps a little over-stated (all bold and exclamation points) given it's mainly a concern for personal usage? Let's either change the 🔸 to a❗or remove the second exclamation point.

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+-  Kinesis Streams' shards each only permit 5 reads per second. If you are using ```n``` shards in a particular stream, and evenly distributing your data across all of them, you will end up with a total of 5 reads per second. This is because a consumer cannot know which shard will contain new data, and will therefore need to check every single one. This means that there is a hard limit on the number of consumers you can have per stream, for any given latency.
+	- If you wish to have 5 consumers all reading data from one Stream with 5 shards, with a maximum permitted latency of 0.5 seconds, you will need to either split your data across two streams, or reduce your latency requirements - with the setup described above, each consumer will need to poll each shard once every 0.5 seconds, meaning each Shard will need to be queried 10 times a second - a value in excess of the maximum.
```
jlevy: Replace the dashes with period and comma, respectively.

Ideally this text could be tightened up a little too.

"Shard" -> "shard"

README.md
```diff
@@ -1289,6 +1291,33 @@ EMR
...
+	- There is a good blog by Brandur, an engineer at Stripe, that discusses the performance and limitations of Kinesis in production [here](https://brandur.org/kinesis-in-production).
```
jlevy: Minor:
- [This blog](https://brandur.org/kinesis-in-production) (by an engineer at Stripe) elaborates on the performance and limitations of Kinesis in production.

README.md
KrishanBhasin: I see your point, and am trying to rewrite this bit to be cleaner.

The idea I'm trying to portray is that this issue is encountered when you have multiple consuming applications, **each** with their own maximum read latency, i.e. it is not the distribution across multiple shards which is the problem, but the fact that multiple different consumers need low latency access to the records within the shards.

PR #200, opened 2016-10-14
Expand the serverless section of the service matrix

chids:
- Providers: Auth0 Webtask and PubNub Blocks


jlevy:
These seem like reasonable additions, so we'll definitely consider them. Would you mind removing the links for now (since we're not linking anyone at the moment) and resolving conflicts?

jlevy:
Also, we'd like to have a sense if these are the top providers; do you have any info or sense of this? We can also discuss on Slack. Thanks for the contribution!

chids:
> Also, we'd like to have a sense if these are the top providers

I'm by no means an expert in the field but my feeling is that apart from the big three (AWS, Microsoft & Google) it might be too early to state if any alternatives such as these are "top providers". However both PubNub and Auth0 are established players in their "traditional" fields and have only recently moved into the serverless space.

> We can also discuss on Slack

As we seem to be in pretty different timezones it might be easier to keep the dialog here? Might also be easier for others to track down the discussion and chime in if it's kept together with the PR?

ThanosBaskous:
Thanks @chids!

PR #222, opened 2016-10-16
Remove redundant word

jbao:


ThanosBaskous:
Thanks for the fix, @jbao!

PR #224, opened 2016-10-17
adding pricing and gotchas to the Lambda section

em0ney:


README.md
```diff
@@ -1048,6 +1048,7 @@ Lambda
 
 -	📒 [Homepage](https://aws.amazon.com/lambda/) ∙ [Developer guide](http://docs.aws.amazon.com/lambda/latest/dg/) ∙ [FAQ](https://aws.amazon.com/lambda/faqs/) ∙ [Pricing](https://aws.amazon.com/lambda/pricing/)
 -	**Lambda** is a relatively new service (launched at end of 2014) that offers a different type of compute abstraction: A user-defined function that can perform a small operation, where AWS manages provisioning and scheduling how it is run.
+-	💸Lambda is free for the first 1 million requests and costs $0.20 for each 1 million requests thereafter. ([See pricing page](https://aws.amazon.com/lambda/pricing/)).
```
jlevy: Let's skip this line. Most of the different services have a free level and then tiered pricing, and I'm not sure we want to commit to maintaining price numbers on everything right in the guide.

README.md
```diff
@@ -1066,6 +1067,8 @@ Lambda
 -	🔸Lambda is a new technology. As of mid 2016, only a few companies are using it for large-scale production applications.
 -	🔸Managing lots of Lambda functions is a workflow challenge, and tooling to manage Lambda deployments is still immature.
 -	🔸AWS’ official workflow around managing function [versioning and aliases](https://docs.aws.amazon.com/lambda/latest/dg/versioning-aliases.html) is painful.
+-	🔸Currently, Lambda does not natively provide for environment variables, which are a common way of storing configuration (see [factor 3 - Config, in the 12-factor app](https://12factor.net/config)). This is currently provided as part of the [Serverless Framework](https://github.com/serverless/serverless)
```
jlevy: I do think a link about this point is helpful, but shortening would help:

> Currently, Lambda does not natively provide for environment variables, which are a [common way](https://12factor.net/config) of storing configuration, so you must rely on additional mechanisms (such as the [Serverless](https://github.com/serverless/serverless) framework).

jlevy: Also, is there any other link about (lack of) environment variable support in Lambda that could be added here?

README.md
chids: I agree with @jlevy but also think that the writing as proposed is too simplified since Lambda is priced on both requests and "GB-seconds" which is a metric for compute time that differs depending on the memory setting of the Lambda.

As an example the current free tier is...

> 1 million requests _and_ 400,000 GB-seconds of compute time per month

...so the amount of actual invocations you can serve in the free tier depends both on the memory setting of the Lambda as well as the execution time, on a per-request basis, of the function.

jlevy: For this PR, let's not add this line at all. If there are counter-intuitive things or gotchas re pricing, let's do a second PR. Re-summarizing the whole pricing schemes isn't a good use of the space here if it's already summarized in the link above.

em0ney:
@jlevy, I implemented your feedback - let me know if there is anything else

jlevy:
LGTM, thanks! (Tiny nit: Missing period at end.) And sadly another merge conflict.

Over to @ThanosBaskous or @ehammond want to do a second topic review?

ThanosBaskous:
LGTM!

em0ney:
Merged upstream and fixed the full-stop (aka period :P) @jlevy

ThanosBaskous:
Thank you for your submission, @em0ney!

PR #229, opened 2016-10-17
Kinesis Streams cleanup & fixes

KrishanBhasin:
Changes made as per @jlevy comments. Further feedback welcome!


README.md
```diff
@@ -1298,25 +1298,29 @@ Kinesis Streams
...
--   Kinesis uses very AWS-specific APIs and terms (e.g. Shards), so you should be aware of the potential future costs of migrating away from it, should you choose to use it.
+-	🚪 Kinesis is most closely compared to [Apache Kafka](https://kafka.apache.org/), an open-source data ingestion solution. It is possible to set up a Kafka cluster hosted on [EC2 instances](#ec2) (or any other VPS), however you are responsible for managing and maintaining both Zookeeper and the Kafka brokers in a highly available configuration. Confluent has a good blog post with their recommendations on how to do this [here](http://www.confluent.io/blog/design-and-deployment-considerations-for-deploying-apache-kafka-on-aws/), which has links on the bottom to several other blogs they have written on the subject.
+-   ⛓ Kinesis uses very AWS-specific APIs, so you should be aware of the potential future costs of migrating away from it, should you choose to use it.
+    -   An application that efficiently uses Kinesis Streams will scale the number of shards up and down based on the required streaming capacity - there is no direct equivalent to this with Apache Kafka.
```
ThanosBaskous: nit: use em dash please (–)

jlevy: Yes, please use em dashes everywhere.

Though actually in this case I think it's better to say:
> An application that efficiently uses Kinesis Streams will scale the number of shards up and down based on the required streaming capacity. (Note there is no direct equivalent to this with Apache Kafka.)

jlevy: Also, this bullet should be one level up (not indented), as it's about a feature, and not about lock-in.

README.md
```diff
@@ -1298,25 +1298,29 @@ Kinesis Streams
...
-	- The KCL uses a DynamoDB table to keep track of which records have been processed by the KCL. This ensures that all records are processed ‘at least once’. It is up to the developer to ensure that the program can handle doubly-processed records.
-	- The KCL also uses DynamoDB to keep track of other KCL ‘workers’. It automatically shares the available Kinesis Shards across all the workers as equally as possible.
+-	The [KCL](https://docs.aws.amazon.com/streams/latest/dev/developing-consumers-with-kcl.html) (Kinesis Client Library) provides a skeleton interface for Java, Node, Python, Ruby and .NET programs to easily consume data from a Kinesis Stream. It provides the outline for 3 basic functions - ```initialize```, ```process-records```, and ```shutdown```. As a developer, all you need to do is set up the config file to point at the correct Kinesis Stream and fill out the provided functions in order to start consuming data from the Stream.
```
jlevy: Checking again: It looks like these are not the exact names, right? https://github.com/awslabs/amazon-kinesis-client/blob/master/src/main/java/com/amazonaws/services/kinesis/multilang/MultiLangProtocol.java
(also there are typographic variations?).

The names don't add much. Can we just combine the last two sentences: "As a developer, you provide functions for initializing, processing records, and shutting down, and a config file to point to the correct Kinesis Stream, in order to start consuming data from the Stream."

README.md
```diff
@@ -1298,25 +1298,29 @@ Kinesis Streams
...
-	- If you wish to have 5 consumers all reading data from one Stream with 5 shards with a maximum permitted latency of 0.5 seconds, you will need to either split your data across two streams or reduce your latency requirements — with the setup described above, each consumer will need to poll each shard once every 0.5 seconds, meaning each Shard will need to be queried 10 times a second — a value in excess of the maximum.
-	- There is a good blog post by Brandur, an engineer at Stripe, which discusses the performance and limitations of Kinesis in production [here](https://brandur.org/kinesis-in-production).
+- 🔸 Kinesis Streams’ shards each only permit [5 reads per second](http://docs.aws.amazon.com/streams/latest/dev/service-sizes-and-limits.html). If you are evenly distrubuting data across many shards, your read limit for the Stream will remain at 5 reads per second on aggregate, as each consuming application will need to check every single shard for new records. This puts a hard limit on the number of different consuming applications possible per Stream for a given maximum read latency.
```
jlevy: Add 🔸⏱ (perf icon too)

README.md
```diff
@@ -1298,25 +1298,29 @@ Kinesis Streams
...
+   - An example: if you wish to have 5 consuming applications reading data from one Stream with ```n``` shards, with a desired maximum latency of 0.5 seconds, each consumer will need to poll each shard once every 0.5 seconds. This means each shard will need to be queried 10 times a second, a value higher than the limit of 5 per second.
```
jlevy: Simpler:
> For example, if you have 5 consuming applications reading data from one Stream with many shards, you cannot have a latency less than 1 second, since each consumer will need to poll each shard every second.

Is that right? If so I think it's clearer.

KrishanBhasin: I like the clarity you add, but I'm wary of saying 'many shards', as it implies that the problem doesn't exist with few shards - the key here is that this limit exists irrespective of the number of shards.

jlevy: OK, how about
> For example, if you have 5 consuming applications reading data from one Stream (with any number of shards), you cannot have a latency less than 1 second, since each consumer will need to poll each shard every second.

jlevy:
LGTM! @ThanosBaskous ?

ThanosBaskous:
LGTM - thanks @KrishanBhasin and @jlevy!

PR #249, opened 2016-10-21
fix typos and some rephrasings

thibaultcha:
* fix typos
* rephrase an awkward sentence (which had a typo too)
* rephrase the `managing multiple regions is complex` sentence which was a bit awkward as well


jlevy:
👍 Thanks for the fixes!

PR #203, opened 2016-10-14
Broke RDS into multiple sections for each database

danhermann:
Broke RDS into multiple sections for each database and added
RDS-specific content for MySQL, Aurora, MariaDB, and SQL Server

PR #251, opened 2016-10-23
Break RDS into separate sections for MySQL, Aurora, and SQL Server

danhermann:
This should replace PR #203.


AlJohri:
not sure if this is expected but I'm seeing some weird encoding issues in the diff

screen shot 2016-10-23 at 3 40 19 pm

danhermann:
I'll take a look at those and see if I can resolve them.

ThanosBaskous:
Thanks @danhermann - I see the same thing as @AlJohri (thanks for calling it out).

ThanosBaskous:
Thanks for fixing those sections, @danhermann - what did you change? Had you copy and pasted into an editor with different encoding or without full unicode support? Unfortunately it looks like the problem still somehow exists with other emoji/unicode chars in the commit (ones in areas that you did not change).

danhermann:
Retrying once more in #255.

danhermann:
@ThanosBaskous, I'm doing this editing on a different machine than my usual dev environment and both its UTF and git setup is different than what I'm accustomed to. Fun times. :)

PR #255, opened 2016-10-24
Break RDS into separate sections for MySQL, Aurora, and SQL Server

danhermann:
Third time's the charm.


ThanosBaskous:
Thanks for sticking with it through the complications, @danhermann, and thanks for this excellent addition!

PR #268, opened 2016-10-26
added tip about default UTC timezone for RDS instances

danhermann:


ThanosBaskous:
LGTM - thank you @danhermann! Bouncing to @jlevy for second stage review.

jlevy:
👍 Keep the great additions coming!

One thought, maybe it's worth adding a parenthetical note about the dangers of non-UTC timezones...

> (❗However, as discussed [earlier](#managing-servers-and-applications), in most situations, it is strongly recommended to use UTC consistently everywhere for all servers and databases.)

Or a broader discussion on this.

danhermann:
@jlevy, I definitely agree that using non-UTC timezones is almost always the wrong thing to do. I tend to think of that as a basic system design principle that's not specific to AWS, but if there's another discussion on it or you'd like something written up about it, I could take a crack at it.

jlevy:
@danhermann Yes, that'd be great! Added issue to track (and commented more there).

PR #278, opened 2016-10-30
add info on non-default cluster param groups

lynnlangit:
similar to info on this topic in RDS, just ran into it with a client (again!) last week, so wanted to add it here


jlevy:
LGTM. Thanks @lynnlangit ! Over to @max-zanko for topic review.

PR #279, opened 2016-10-30
Add Readme s3 gotcha: authentication

rjhintz:
This PR is not associated with an issue. No specific discussion except on Slack.

S3 CLI/SDK calls in new regions require the latest Authentication Signature version. Older regions can use latest or previous versions. This can result in calls in old regions working, but not in new regions.

Edit: didn't squash out my previous merges. Apologies if this is bad style for this repo.


README.md
```diff
@@ -699,6 +699,7 @@ S3
...
 -	❗When configuring ACLs on who can access the bucket and contents, a predefined group exists called **[Authenticated Users](http://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html)**. This group is often used, incorrectly, to restrict S3 resource access to authenticated users of the owning account. If granted, the AuthenticatedUsers group will allow S3 resource access to **all authenticated users, across all AWS accounts**. A typical use case of this ACL is used in conjuction with the [requester pays](http://docs.aws.amazon.com/AmazonS3/latest/dev/RequesterPaysBuckets.html) functionality of S3.
+- ❗If a S3 file operation using CLI or SDK doesn't work in one region but works correctly in another region, make sure you are using the latest [Authentication Signature](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html). See also [Specifying Signature Version in Request Authentication](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingAWSSDK.html#specify-signature-version).
```
jlevy: How about a couple tweaks to explain the rationale. Does this look right?

> - 🔸**S3 authentication versions and regions:** In newer regions, S3 [only supports Version 4 authentication](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingAWSSDK.html#specify-signature-version). If a S3 file operation using CLI or SDK doesn't work in one region but works correctly in another region, make sure you are using the latest [authentication signature](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html).

(also use 🔸 instead of ❗ as consequences of the gotcha is moderate, rather than severe like a major security risk, outage, etc)

jlevy:
+ @max-zanko for topic review.

README.md
rjhintz: @jlevy   I left out the version number to allow the point to carry forward as new signature versions are introduced, so a bit of future proofing. I don't have an objection to changing if that makes sense, though.

Also, fine with me to change the indicator to less than severe, though the CLI/SDK call won't work without it, which is why I chose the more severe indicator. Again, no problem with changing as you suggest.

rjhintz:
I'm fine with anyone with experience in the project modifying the language to be aligned with the rest of the project. I'm new here, so don't want to be prescriptive.

README.md
max-zanko: Minor nit: If *an* S3 file...
Otherwise LGTM after changes suggested by @jlevy.

rjhintz:
@jlevy @max-zanko Ok, after a lot of git fun, I think this revision incorporates the changes that were requested.

jlevy:
👍 Thanks!

For reference, it's cleaner to avoid committing merge conflicts or extra commits. If you do, you can always [merge --squash or rebase -i](http://stackoverflow.com/questions/2427238/in-git-what-is-the-difference-between-merge-squash-and-rebase)

rjhintz:
Yes, sorry for the mess. I won't trouble you with a lame excuse.

PR #281, opened 2016-10-30
Added ALB and CLB  gotchas about 2048 bit ssl cert

Bizzelicious:


jlevy:
+ @ThanosBaskous for topic review at once.

ThanosBaskous:
The content here is good -  thanks @Bizzelicious. I agree with @jlevy that it should be consolidated into the general load balancer section. Please see phrasing suggestion in #281 as well.

Bizzelicious:
You can now have 4096 certs on CLBs and ALBs. However, CloudFront doesn't support it yet so #282 is still open.

jlevy:
OK, thanks @Bizzelicious for following up!

PR #283, opened 2016-10-31
Clarify gotcha severities.

jlevy:
It seems helpful to formalize this distinction — in other words, it's based on the severity of the bad consequences, not whether something works or not. Sound good? Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,8 @@ Legend
 
 -	📒 Marks standard/official AWS pages and docs
 -	🔹 Important or often overlooked tip
--	❗ Gotcha or warning (where risks or time or resource costs are significant)
--	🔸 Limitation or quirk (where it’s not quite so bad)
+-	❗ “Serious” gotcha (used where risks or time or resource costs are significant: critical security risks, mistakes with significant financial cost, or poor architectural choices that are fundamentally difficult to correct)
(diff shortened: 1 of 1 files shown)


rjhintz:
Maybe two cases, though YMMV:

- Warning: Things break/don't work and you can't figure out why by searching StackOverflow, the AWS docs, or og-aws with simple, intuitive keywords
- Caution: Things don't work, but the cause is in the docs

I realize this is not backwards compatible with the current marking scheme, but reflects how I find problems IRL. I don't have enough context with the project to want to go against a dissenting consensus, though.

ThanosBaskous:
Thanks for the idea @rjhintz - I'm not sure I see the value in connecting the scheme to availability of the information in other docs or on  other sites, though. I'm open to discussing further here or on Slack, though.

@jlevy the PR LGTM. 👍

jlevy:
Agree, @rjhintz those considerations matter, and are a key reason this guide has value — that it contains things you can't find on SO/AWS docs/Google.

However, this PR is about how we annotate things by severity. Ideally, a reader looking to find really serious gotchas that are critical to know should be able to skim/search to find those in this guide. Let's go with this, as it's our current practice.

rjhintz:
I'm fine with this outcome.

PR #292, opened 2016-11-04
added redshift distribution column type info

lynnlangit:
affects query performance


README.md
```diff
@@ -1376,7 +1376,7 @@ Redshift
...
 -	A compound sort key is made up of all columns listed in the sort key definition. It is most useful when you have queries with operations using prefix of the sortkey.
 -	An interleaved sort key on the other hand gives equal weight to each column or a subset of columns in the sort key. So if you don't know ahead of time which column you want to choose for sorting and filtering, this is a much better choice than the compound key.[Here](https://aws.amazon.com/blogs/aws/quickly-filter-data-in-amazon-redshift-using-interleaved-sorting/) is an example using interleaved sort key.
-
+-	Redshift includes a ability to condigure a [distribution key](http://docs.aws.amazon.com/redshift/latest/dg/c_choosing_dist_sort.html) for each table. The distribution key that you create on a table sets the type of physical data distribution for that table in your cluster and can have a significant impact on query performance.  There are three possible distribution key settings -- even (which is the default), key or all.  Use 'key' to collocate join key columns for tables which are joined in queries.  Use 'all' to place the data in small-sized tables on all cluster nodes.
```
max-zanko: This change overlaps a lot with the http://docs.aws.amazon.com/redshift/latest/dg/c_choosing_dist_sort.html documentation article, which is pretty concise.
Should we rephrase it so it is an entry point to the official documentation, something like:
"Redshift is a distributed system. Choosing the right data distribution strategy is crucial to the adequate query performance. The primary way to control the distribution is to set the the right distribution key and the distribution style for the table. Make sure you choose the right [distribution style](http://docs.aws.amazon.com/redshift/latest/dg/c_choosing_dist_sort.html) based on your use cases".

jlevy: I'll defer to whatever you guys think for actual recommendations, but stylistically if this is a notable performance gotcha, then let's indicate this, and also suggest more boldfacing of the terms and using AWS' terminology, something like this:

- 	🔸⏱ **Distribution strategies:** Since data in Redshift is physically distributed among nodes, choosing the right data **distribution key** and [**distribution style**](http://docs.aws.amazon.com/redshift/latest/dg/c_choosing_dist_sort.html) is crucial for adequate query performance. There are three possible distribution style settings settings — **EVEN** (the default), **KEY**, or **ALL**.  Use KEY to collocate join key columns for tables which are joined in queries.  Use ALL to place the data in small-sized tables on all cluster nodes.

lynnlangit:
anything else to do here?

jlevy:
LGTM. @max-zanko can you merge if you're good? Thanks @lynnlangit !

Issue #130, opened 2016-10-12
Legend icons should have tooltip


jjcf89:
just incase you've forgotten what the bird means :)

ThanosBaskous:
👍

bfred-it:
Tooltips can be added via anchor:

```md
[🚪](#legend "Alternative non-AWS options")
```

Also linking back to the Legend section would make it usable on mobile (where tooltips don't exist).

danhermann:
+1. I can never remember what the different icons mean.

rjhintz:
I did a "down payment" PR on this to try to bring things up to date. There may be a better way to keep icons current with tooltips going forward that eludes me now.

Icons may be an ongoing maintenance challenge.

jlevy:
This definitely would be really helpful if we can implement it. But with current workflow, it will be really hard to maintain or enforce. I think we may have to hold off on it until we develop a new workflow (such as a post-processing script or a different site hosting it altogether).

PR #301, opened 2016-11-08
Resolve Issue #130 tooltips

rjhintz:
Added tooltips, though ongoing maintenance may be a nightmare. 

Reviewing is tough because of the size of the diff. I tried to figure out an automated way, but failed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,7 +531,7 @@ We cover security basics first, since configuring user accounts is something you
 
 ### Security and IAM Basics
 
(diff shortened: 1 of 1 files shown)


rjhintz:
I don't have objections if this is felt to be too big to digest and either is broken up into multiples PRs by icon to keep the diff manageable, or even deferred while the workflow is rationalized to make sure future icons get tooltips.

jlevy:
As discussed in Slack and #130 we'll need to revisit this later. Merging and maintaining this with current workflow isn't something we can expect to sustain. Thanks again!

rjhintz:
I agree that this is an appropriate resolution.

PR #312, opened 2016-11-13
Remove note about nginx name resolution issue

ram-nadella:
One of the Load Balancer gotchas mentions nginx causing issues with load balancers because name resolution happens only at start-up. This used to be the case with older version of nginx but is not longer correct as mentioned in the blog post linked to in the note and in the official docs (http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver).


README.md
```diff
@@ -924,7 +924,7 @@ Load Balancers
...
--	❗Some web clients or reverse proxies cache DNS lookups for a long time, which is problematic for CLBs and ALBs, since they change their IPs. This means after a few minutes, hours, or days, your client will stop working, unless you disable DNS caching. Watch out for [Java’s settings](http://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html) and be sure to [adjust them properly](http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/java-dg-jvm-ttl.html). Another example is nginx as a reverse proxy, which [resolves backends only at start-up](https://www.jethrocarr.com/2013/11/02/nginx-reverse-proxies-and-dns-resolution/).
+-	❗Some web clients or reverse proxies cache DNS lookups for a long time, which is problematic for CLBs and ALBs, since they change their IPs. This means after a few minutes, hours, or days, your client will stop working, unless you disable DNS caching. Watch out for [Java’s settings](http://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html) and be sure to [adjust them properly](http://docs.aws.amazon.com/AWSSdkDocsJava/latest/DeveloperGuide/java-dg-jvm-ttl.html).
```
jlevy: Hm. I think omitting it entirely means people might not be aware of this kind of issue, and many ppl are out of date on systems like nginx. It's also good for those of us who remember the old functionality to know it's now changed. How about this:

> Another example is that older versions nginx [used to resolve reverse proxy backends only at start-up](https://www.jethrocarr.com/2013/11/02/nginx-reverse-proxies-and-dns-resolution/), though this [is now fixed](http://nginx.org/en/docs/http/ngx_http_core_module.html#resolver).

jlevy:
Adding @ThanosBaskous who may have another take on this.

ThanosBaskous:
It's good to keep this up to date - thanks for the submission @ram-nadella! I agree with @jlevy that it's worth calling out the old behavior both as an example and in case people are still running it (many are). @ram-nadella - would you mind incorporating (something like) @jlevy's suggestion?

README.md
ThanosBaskous: I agree - one missed word in your example:  "older versions nginx" -> "older versions of nginx".

ram-nadella:
@jlevy that's a good suggestion.

I've got some news though: before committing the updated text I did a review of the situation mentioned in the blog post and I realized that the official docs I point to only mention the `resolve` directive but does not talk about a change in behaviour regarding how backend/upstream server addresses are resolved.

So I went looking for more info and it turns out this is still an issue, but there is a [clever way to write config](https://tenzer.dk/nginx-with-dynamic-upstreams/) to force nginx to resolve backends dynamically instead of only on start-up / config reload.

I am going to edit the line to something like this:

Another example is nginx as a reverse proxy, which [resolves backends only at start-up](https://www.jethrocarr.com/2013/11/02/nginx-reverse-proxies-and-dns-resolution/), although there is [a way to get around this](https://tenzer.dk/nginx-with-dynamic-upstreams/).

Thoughts?

jlevy:
@ram-nadella thanks for looking into it more closely.

That change looks good. Minor suggestion:

> Another example is nginx as a reverse proxy, which [normally resolves backends only at start-up](https://www.jethrocarr.com/2013/11/02/nginx-reverse-proxies-and-dns-resolution/) (although there is [a way to get around this](https://tenzer.dk/nginx-with-dynamic-upstreams/)).

Ping me when the PR is ready. Thanks!

ThanosBaskous:
@ram-nadella thank you for being thorough in your research!

ram-nadella:
Ran into some branch / PR issues, so I ended up creating a [new PR](https://github.com/open-guides/og-aws/pull/321). That other PR doesn't quite look right either (includes the commits since my clone), think it had to do with how my fork was brought up to date with upstream (did it via the Github UI).

This is the final [commit](https://github.com/ram-nadella/og-aws/commit/9f9549d952d4d5a8cccc3bc7d2070d7940a623a9) based on our discussion here.

PR #314, opened 2016-11-14
Remove word repetitions

danbower:
Thanks for the great guide! Just noticed a couple minor things whilst reading through so thought I'd do a quick pull request. The repetitions were "where where" and "settings settings".


jlevy:
👍 Thanks for the fixes!

PR #321, opened 2016-11-20
Update note about nginx gotcha in the CLB section

ram-nadella:
More discussion about this happened in a different [pull request](https://github.com/open-guides/og-aws/pull/312).


jlevy:
👍 Thanks @ram-nadella !

PR #336, opened 2016-11-27
Updated API Gateway gotchas

russelldavies:
Added note about lack of binary data support.


russelldavies:
Oh very cool, they seem to be making a lot of changes at the moment. I'll add it as a tip instead.

jlevy:
👍 Yep, looks good!

PR #356, opened 2016-12-14
Fix incorrect EFS information

brwilson:


README.md
```diff
@@ -943,8 +943,8 @@ EFS
 -	🔸 A number of NFSv4.1 features are [not supported](http://docs.aws.amazon.com/efs/latest/ug/nfs4-unsupported-features.html) and there are some [limits](http://docs.aws.amazon.com/efs/latest/ug/limits.html) to the service.
 -	🔸 As of 2016-11, EFS does not offer disk level encryption, though it is on the roadmap.
 -	❗ Some applications, like SQLite and IPython, [might not work properly](https://sqlite.org/faq.html#q5) on EFS when accessed from multiple clients. This is because lock upgrades and downgrades are [not supported](http://docs.aws.amazon.com/efs/latest/ug/nfs4-unsupported-features.html). There might be [workarounds](https://github.com/ipython/ipython/issues/882) for some issues.
--	🔸 Mounting EFS over VPN connection, VPC peering, or AWS Direct Connect is not supported.
--	🔸 Using an EFS volume on Windows is not supported, apparently due to Microsoft implementing NFS differently.
+-	🔸 Mounting EFS over VPN connection, VPC peering, ~~or AWS Direct Connect~~ is not supported. EFS [does support](https://aws.amazon.com/efs/faq/#on-premises) mounting over Direct Connect.
```
jlevy: If it's incorrect, let's just remove the strikethrough text altogether:

> 🔸 Mounting EFS over VPN connection or VPC peering not supported. (EFS [does support](https://aws.amazon.com/efs/faq/#on-premises) mounting over Direct Connect.)

README.md
```diff
@@ -943,8 +943,8 @@ EFS
...
+-	🔸 Using an EFS volume on Windows is not supported because while the server service supports NFS v3 and v4.1, the client only supports NFS v3.
```
jlevy: Glad this point is clarified, but still, from what's said here it's not clear why that doesn't work? Can the server not detect the version? Is there a link you can add?

brwilson: Sorry, that wording still wasn't very clear. Windows can act as an NFS server or an NFS client. When used as an NFS server, it can present shares using NFS 2, 3, and 4.1 (not 4). When used as an NFS client, it can only mount NFS v3 (and maybe v2?) shares.

The NFS client docs are pretty sparse, but if you look at the flags you'll see there's no way to specify an NFS version:
https://technet.microsoft.com/en-us/library/cc754350(v=ws.11).aspx

Also, there was a post in the AWS forums where one of the AWS folks claimed the client was NFSv3 only:
https://forums.aws.amazon.com/thread.jspa?threadID=234444

I'm fine with reverting the edit if you're not comfortable with that level of evidence—regardless of the reason, Windows clients won't be able to mount EFS file systems, and that's the important piece of information.

README.md
```diff
@@ -943,8 +943,8 @@ EFS
 -	🔸 A number of NFSv4.1 features are [not supported](http://docs.aws.amazon.com/efs/latest/ug/nfs4-unsupported-features.html) and there are some [limits](http://docs.aws.amazon.com/efs/latest/ug/limits.html) to the service.
 -	🔸 As of 2016-11, EFS does not offer disk level encryption, though it is on the roadmap.
 -	❗ Some applications, like SQLite and IPython, [might not work properly](https://sqlite.org/faq.html#q5) on EFS when accessed from multiple clients. This is because lock upgrades and downgrades are [not supported](http://docs.aws.amazon.com/efs/latest/ug/nfs4-unsupported-features.html). There might be [workarounds](https://github.com/ipython/ipython/issues/882) for some issues.
--	🔸 Mounting EFS over VPN connection, VPC peering, or AWS Direct Connect is not supported.
--	🔸 Using an EFS volume on Windows is not supported, apparently due to Microsoft implementing NFS differently.
+-	🔸 Mounting EFS over a VPN connection or VPC peering is not supported. EFS [does support](https://aws.amazon.com/efs/faq/#on-premises) mounting over Direct Connect.
```
bgdnlp: Propose rephrase:
-	🔸 An EFS file system can be [mounted on premises](https://aws.amazon.com/efs/faq/#on-premises) over Direct Connect.
-	🔸 An EFS file system can NOT be mounted over VPC peering or VPN, even if the VPN is running on top of Direct Connect.

The reason why it's not working over VPN is that the mount targets are not directly subject to the subnet routing table, it's done before that by DC. So the packets must not be encrypted. Source, Amazon support.

README.md
```diff
@@ -943,8 +943,8 @@ EFS
...
+-	🔸 Using an EFS volume on Windows is not supported because the Window NFS client does not support NFS v4 or v4.1.
```
bgdnlp: My wording leaves a lot to be desired, I agree. I don't have the practical experience to better phrase this point, but:
1. https://technet.microsoft.com/en-us/library/jj592688(v=ws.11).aspx#Anchor_2 clearly states that NFSv4.1 is supported on Windows, so saying that Windows NFS client doesn't supports NFSv4.1 is false. However, https://technet.microsoft.com/en-us/library/jj574143.aspx#Anchor_2 (and other sources) say that Windows does not support NFSv4, they skipped straight to v4.1. EFS is 4.1 though, so non-support for v4 **should** not be relevant.
2. A more technical explanation about what's different on Windows can be found on the [Unsupported features](http://docs.aws.amazon.com/efs/latest/ug/nfs4-unsupported-features.html) page (linked a few lines above):
> NFS supports the concept of a share deny, primarily used by Windows clients for users to deny others access to a particular file that has been opened. Amazon EFS does not support this, and returns the NFS error NFS4ERR_NOTSUPP for any OPEN commands specifying a share deny value other than OPEN4_SHARE_DENY_NONE. Linux NFS clients do not use anything other than OPEN4_SHARE_DENY_NONE.

3. My understanding is that there are non-official clients that might be able to mount EFS, but, again, I have no practical knowledge of that.

4. I wouldn't argue against removing the "Microsoft implementing NFS differently" comment completely, leaving just the first part. It is vague, though technically true.

jlevy:
Thanks for the review @bgdnlp ! 
@brwilson You can take bgdnlp's feedback and incorporate as best as you can? The two of you know more than I do on this. At this point, if you're both happy, I am. :)

README.md
brwilson: Works for me.

README.md
brwilson: In point 1, the page you linked is referencing the server component, not the client. I spun up a Server 2016 instance and tried to mount a v4.1 share while running a packet capture on the Linux server hosting the share. The Windows client only tried to mount via v3. I was able to mount the share via NFSv3 after enabling v3 on the Linux server. Having said that, I can't point to any documentation that clearly states the client is v3 only, so I'm happy to go with your suggestion in point 4. It's certainly possible I screwed up my testing :)

README.md
```diff
@@ -943,8 +943,10 @@ EFS
 -	🔸 A number of NFSv4.1 features are [not supported](http://docs.aws.amazon.com/efs/latest/ug/nfs4-unsupported-features.html) and there are some [limits](http://docs.aws.amazon.com/efs/latest/ug/limits.html) to the service.
 -	🔸 As of 2016-11, EFS does not offer disk level encryption, though it is on the roadmap.
 -	❗ Some applications, like SQLite and IPython, [might not work properly](https://sqlite.org/faq.html#q5) on EFS when accessed from multiple clients. This is because lock upgrades and downgrades are [not supported](http://docs.aws.amazon.com/efs/latest/ug/nfs4-unsupported-features.html). There might be [workarounds](https://github.com/ipython/ipython/issues/882) for some issues.
--	🔸 Mounting EFS over VPN connection, VPC peering, or AWS Direct Connect is not supported.
--	🔸 Using an EFS volume on Windows is not supported, apparently due to Microsoft implementing NFS differently.
+-	🔸 Mounting EFS over a VPN connection or VPC peering is not supported. EFS [does support](https://aws.amazon.com/efs/faq/#on-premises) mounting over Direct Connect.
+-	🔸 An EFS file system can be mounted on premises over Direct Connect.
```
bgdnlp: Last one, I should have marked it as code in my previous comment. Can you please add the FAQ link that you pointed to originally back? As in:
`-	🔸 An EFS file system [can be mounted on premises](https://aws.amazon.com/efs/faq/#on-premises) over Direct Connect.`

README.md
bgdnlp: No, you're right, searching Google for it turns up only 3rd party clients. I guess that's what Amazon are talking about. Which makes my version not only vague, but plain wrong. Happy with the removal, thank you for pointing it out.

bgdnlp:
@brwilson, just a friendly reminder about the link above. Or is it OK if this is merged and I change the link after that? Because it's good info, don't want to hold it off just for that.

brwilson:
Sorry for the delay! Feel free to merge/edit if there are any other changes you'd like to make.

jlevy:
LGTM. Thanks! Over to @ThanosBaskous to review/merge.

jlevy:
Merging now for expediency, as we've topic editor review already.

PR #357, opened 2016-12-15
replaced Google Brillo with Android Things

lynnlangit:
Google announced that Brillo was being replaced by 'Android Things' as their IoT OS


lynnlangit:
Anything else for me to do here?

jlevy:
Thanks! (And sorry for the delay!)

PR #362, opened 2016-12-28
Added: Snowmobile

QuinnyPig:
Included a line about Snowmobile. Few people will need this.


jlevy:
👍 A little flippant but I'm OK with if @ThanosBaskous is. :)

ThanosBaskous:
Hah. I like it. Thanks @QuinnyPig!

PR #366, opened 2017-01-07
added info about using your own CA for IoT

lynnlangit:
I have found this feature to be in high demand from my customers, so wanted to add a link to it here


jlevy:
👍 Looks good! Thanks!

PR #367, opened 2017-01-09
Add note about EIP migration between Classic and VPC.

richadams:
Announced here: https://aws.amazon.com/about-aws/whats-new/2017/01/announcing-elastic-ip-address-migration-from-ec2-classic-to-ec2-vpc/


jlevy:
👍 Thanks!

PR #374, opened 2017-01-16
Remove one EFS gitcha that was present twice in different forms

bgdnlp:
Fixes the extra line merged in https://github.com/open-guides/og-aws/pull/356


jlevy:
👍 LGTM. Thanks!

PR #375, opened 2017-01-16
Remove obsolete gotcha about EFS mountpoints

bgdnlp:
Gotcha no longer applicable, AWS implemented a single mount point across AZs. The "old" way still works, but there's no reason to do it like that.


jlevy:
LGTM. Thanks @bgdnlp !

PR #381, opened 2017-02-13
Added Google products to comparison table

danielehrlich:
- Added in [Firebase Authentication](https://firebase.google.com/docs/auth/) to the "User authentication" section of the comparison table.
- Added in [Cloud Speech API](https://cloud.google.com/speech/) to the "Text-to-speech engine in the cloud".


danielehrlich:
Hey All:

I've worked with AWS for the last couple of years, but this is my first commit to this repo. I made sure to include links, but if there is anything else I need to add in terms of documentation or further editing please just let me know.

lukedeo:
Hi @danielehrlich, thanks for the contribution! Minor nit, Cloud Speech technically does speech-to-text - want to add that to the row above, same column? (I.e., in addition to Google's Natural Language API)

danielehrlich:
Ahh good catch and just updated!

jlevy:
👍 Thanks @danielehrlich ! LGTM.

PR #383, opened 2017-02-15
Added Oracle Cloud to comparison chart

danielehrlich:
Hi All:

I added in an Oracle Cloud comparison chart. I'm a Solutions Engineer at Oracle, so that's where my familiarity comes from. 

Here is the main website for the [PaaS Services](https://cloud.oracle.com/en_US/paas).

Here is the main website for the [IaaS Services](https://cloud.oracle.com/en_US/iaas).

Later on I'd like to add in IBM's Cloud, since I work with them a good amount too.


danielehrlich:
Hey All:

I haven't seen anyone comment on this post. My end goal would be to add Oracle Cloud and IBM Cloud to this repo. 

Something like this repo: https://ilyas-it83.github.io/CloudComparer/

I work at Oracle, and I know both Oracle and IBM are historically not good at easy documentation or having easy developer resources (that is changing somewhat though, just type in "www.developer.[ORACLE or IBM].com").  Even though their percentage counts of cloud users are low (4% for each platform if we don't count SaaS?), I think having them in this repo is important.

ThanosBaskous:
Hi @danielehrlich, thank you for your submission. We are trying to keep the list of public cloud options down to the most important 4 or 5 + open source in order to keep the table manageable and the guide navigable, and I don't think that either Oracle or IBM's offerings are large or significant enough to squeeze out any of the others.

If you disagree with the assessment of size or significance of Oracle and IBM offerings relative to the others in the list, I'd be happy to review any information that you submit to support that.

I should also note that we definitely aren't opposed to mentioning the existence of Oracle and IBM's offerings - I'd be supportive of a PR that added a note about the top 5-10 AWS alternatives (of which those two would be a part).

cc @jlevy

danielehrlich:
Hey Thanos:

That all sounds fine and looking over the issue tracker now I see you guys have had to decide things like this before as you try to keep the guide still AWS focused (e.g. I think someone submitted Centurylink Cloud a while ago??). 

I actually wound up just cloning a copy of the Matrix and posting it on my Github (with credit to Joshua Levy & You All!) . I think that's more appropriate because of the reason mentioned above as well as the fact that I'm an Oracle employee. So I'm able to explicitly mention that for Oracle's social media policy and also the community at large.

PR #387, opened 2017-03-04
Added in section for File Share and Sync

danielehrlich:
The official name for this category is Enterprise File Share and Sync (EFSS). Was not sure where to place it overall so added it at the bottom. All of these services have pretty good APIs, I think there are at least a few cases where devs could be more innovative by using this for their storage instead of typical object storage.


ThanosBaskous:
lgtm, thanks for your contribution @danielehrlich!

danielehrlich:
Thanks @ThanosBaskous :)

PR #390, opened 2017-03-07
Remove the limitation tip about NFSv4 file locking

bgdnlp:
As per https://aws.amazon.com/about-aws/whats-new/2017/03/amazon-elastic-file-system-amazon-efs-now-supports-nfsv4-lock-upgrading-and-downgrading/


jlevy:
LGTM. Thanks @bgdnlp !
@ThanosBaskous I'll merge since bgdnlp is editor but feel free to post-hoc review. :)

PR #396, opened 2017-03-20
adding AWS Batch to list of services

lynnlangit:


ThanosBaskous:
Thank you @lynnlangit!

PR #398, opened 2017-03-24
Elastic's own cloud offering is important

sunygit:
This should be on the list: https://www.elastic.co/de/cloud/as-a-service (formerly called "Found", a company that was bought by elastic in 2015). You get the latest versions, Kibana, custom plugins and scripts and dictionaries. Also, you can access the commercial x-pack products (Shield / Marvel / Watcher / Graph / Reporting), also in the free 14 days version.
I't possible to add Elastic Cloud onto your AWS bill now, it's in the marketplace
https://aws.amazon.com/marketplace/pp/B01N6YCISK
In case you wonder, I'm not from the company elastic.


jlevy:
Cool, thanks. This seems like a helpful and notable addition to me. If @ThanosBaskous or another editor agrees, let's merge. (Marking as "neutrality" as track vendor links like this.)

ThanosBaskous:
I agree. Thanks @sunygit!

PR #414, opened 2017-04-13
Added Last Week in AWS to further reading section

QuinnyPig:
Note that this requires a bit of scrutiny; I'm the editor of the newsletter in question, so my bias is rather clear!


ThanosBaskous:
Thanks for calling out your association @QuinnyPig - I think that a link out to "Last Week in AWS" is a great addition to the guide. I'm supportive of this change - will wait for one more editor to +1 before one of us merges.

jlevy:
I'm in favor too — it's been well received by folks here, and seems there's no other newsletter like it.

jlevy:
Given we usually leave "neutrality" ones open a little longer in case there are comments, I'll wait until Mon to merge this.

QuinnyPig:
@jlevy It's been two Mondays. :)

ThanosBaskous:
👍 I think that we've provided enough time for comment on this.

PR #417, opened 2017-04-13
Never suggest us-west-1; it's expensive and massively capacity constrained.

QuinnyPig:


ThanosBaskous:
Never is a strong word - there are good reasons to be in us-west-1. For example, when you need low latencies to ad exchanges, which themselves have sprung up in and around us-west-1. I'd prefer to add more nuance to the recommendation rather than remove us-west-1 altogether. Thoughts?

bgdnlp:
I'd scrap suggesting a region completely. Suggest considering latency (don't set up in US if your operations are in Europe) and pricing (some regions are more expensive than others, even in similar locations, for example us-west-1 vs us-west-2) and let the reader do their own research. Any region suggestion might be invalid in 6 months if Amazon decides so anyway.

ThanosBaskous:
Good suggestions, @bgdnlp, I'd be on board with framing it that way as well.

QuinnyPig:
Reframed; what do you think now?

ThanosBaskous:
That looks great to me - thanks @QuinnyPig!

Issue #385, opened 2017-03-02
Mention an AZ is separate physical infra


jlevy:
From discussion — let's mention/link this as it's a common question.

thanos [9:43 AM] 
From https://www.youtube.com/watch?v=JIQETrFC_SQ&feature=youtu.be&t=1428
"An availability zone is 100% always without ever an exception, a completely independent building. It is a different datacenter."

DavidTPate:
It would be good to mention also that there is low latency between AZs. This was taken from the docs:

> Availability Zones are physically separate, such that even extremely uncommon disasters such as fires, tornadoes or flooding would only affect a single Availability Zone.
> Although Availability Zones within a Region are geographically isolated from each other, they have direct low latency network connectivity between them.

PR #422, opened 2017-04-15
Clarify Availability Zone separation

QuinnyPig:
Address Issue #385


jlevy:
Good addition. This is commonly not understood; can we add a link to https://aws.amazon.com/about-aws/global-infrastructure/ (or anything better) where it says they are "housed in separate facilities"?

ThanosBaskous:
Another reference option (from #385) is https://www.youtube.com/watch?v=JIQETrFC_SQ&feature=youtu.be&t=1428

QuinnyPig:
Best quote I've found so far is "Availability Zones are distinct locations within a region that are engineered to be isolated from failures in other Availability Zones and provide inexpensive, low latency network connectivity to other Availability Zones in the same region." They're cagy about it in writing, and I don't like "here's a YouTube video to watch" as a citation model. :-/

ThanosBaskous:
The video form factor isn't great, but the [source](https://www.linkedin.com/in/jameshamilton4/) is pretty authoritative.

QuinnyPig:
Huh, I kinda dig what https://blog.rackspace.com/aws-101-regions-availability-zones has to say on the matter:

> An Availability Zone is a logical data center in a Region available for use by any AWS customer. Each zone in a Region has redundant and separate power, networking and connectivity to reduce the likelihood of two zones failing simultaneously. A common misconception is that a single zone equals a single data center. In fact, each zone is backed by one or more physical data centers, with the largest backed by five.

Given their enormous partner status, does that help?

ThanosBaskous:
Good find. I'm fine with that as a citation. @jlevy?

jlevy:
Cool. Let's link both! While we're add it did a little wordsmithing/formatting below — see if that looks good.

Side note on referencing videos: Generally we should reference whatever is most authoritative, regardless of form factor, which is why I think we also include the Hamilton video. It also has more contextual info. Videos are annoying to watch if you're in a hurry, but that's exactly why we are summarizing key points for people. :)

---

> **Regions and Availability Zones:** AWS allows deployments in [**regions**](https://docs.aws.amazon.com/general/latest/gr/rande.html), which are isolated geographic locations that help you reduce latency or offer additional redundancy.
Regions contain [**Availability Zones**](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html) (AZs),  which are typically the first tool of choice for [high availability](#high-availability). AZs are [physically separate from one another](https://www.youtube.com/watch?v=JIQETrFC_SQ&feature=youtu.be&t=1428), even within the same region, and [may span multiple physical data centers](https://blog.rackspace.com/aws-101-regions-availability-zones). While AZs are connected via low latency links, natural disasters afflicting one AZ should not affect others.

QuinnyPig:
@jlevy Incorporated. Good to go?

ThanosBaskous:
lgtm @QuinnyPig 👍

ThanosBaskous:
Thanks for staying on top of this one @QuinnyPig!

Issue #340, opened 2016-11-30
Mention new Organizations support


jlevy:
https://aws.amazon.com/about-aws/whats-new/2016/11/announcing-aws-organizations-now-in-preview/

In preview but we should mention its existence.

QuinnyPig:
Mentioned; closing issue.

PR #424, opened 2017-04-15
Added Organizations

QuinnyPig:
Addresses #340


jlevy:
LGTM. Seems great but I'll also ask @ThanosBaskous as there might be other items we should add at the same time?

ThanosBaskous:
LGTM as well - I think that additional Organizations-related can be added in subsequent PRs. Thanks @QuinnyPig

PR #421, opened 2017-04-15
Fixup EFS Wording

QuinnyPig:
Fixed some word choices in the EFS section, renamed File Store to Object Store in the service matrix

PR #426, opened 2017-04-17
Fixing cost misstatement

QuinnyPig:
I screwed up on that one! Credit to @bgdnlp for spotting it.


jlevy:
👍 LGTM. Already discussed in #421 so merging.

PR #425, opened 2017-04-17
Add OpsWorks section.

ClydeMachine:
Create basic OpsWorks section with some general details and gotchas. It's a cleaned up version of previous PR [#348](https://github.com/open-guides/og-aws/pull/348). Fixes #123.

PR #428, opened 2017-04-19
Update information regarding OpsWorks Stacks and Chef Automate region…

ClydeMachine:
… availability.

Just taking care of a couple lingering things noted in the #425 PR. I didn't include coverage of Saltstack as I am not familiar enough with it to make a recommendation/compare features.


ThanosBaskous:
Great! Thanks for addressing that feedback and for doing the additional research, @ClydeMachine.

PR #436, opened 2017-04-30
Adds CloudWatch security gotcha

zekizeki:
* 🔸CloudWatch does not have a [VPC Endpoint](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-endpoints.html). If you have a security requirement to lockdown outbound traffic from your VPC then you may want to look at alternative logging solutions or use [DNS filtering](https://aws.amazon.com/blogs/security/how-to-add-dns-filtering-to-your-nat-instance-with-squid/) to control outbound traffic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -846,6 +846,7 @@ CloudWatch
 * 🔸Notifications you receive from alarms will not have any contextual detail; they have only the specifics of the threshold, alarm state, and timing.
 * 🔸Minimum granularity in CloudWatch is 1 minute. That means that multiple values of a metric that are pushed to CloudWatch within the same minute are aggregated into minimum, maximum, average and total (sum) per minute.
 * 🔸Data about metrics is kept in CloudWatch [for 15 months](https://aws.amazon.com/blogs/aws/amazon-cloudwatch-update-extended-metrics-retention-user-interface-update/), starting November 2016 (used to be 14 days). Minimum granularity increases after 15 days.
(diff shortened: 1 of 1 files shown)


QuinnyPig:
Thanks for the contribution!

I'm on the fence about this one-- you're absolutely correct in your assessment. The challenge I've got is that this applies to every service except S3 and DynamoDB (in preview). 

Does generalizing it and making it a larger point beyond CloudWatch make sense, do you think?

ThanosBaskous:
Thank you, @zekizeki.

I agree with @QuinnyPig's thinking on this. Noting the absence of VPC endpoints for CloudWatch seems like it would be likely to lead readers to believe that that they are there for other services, which is (mostly) not the case.

I'd be very supportive of a modified PR that notes the general lack of VPC endpoints (besides the ones that @QuinnyPig mentioned) - the "VPC and Network Security Gotchas and Limitations" section seems like the right place to me, but I'm open to other ideas.

QuinnyPig:
As I quoted a couple of weeks ago in https://lastweekinaws.com, "Two years after AWS debuted VPC endpoints for S3, a second service now offers them in a limited preview. At this rate, Amazon's rollout of VPC endpoints for all its service offerings should be completed roughly three weeks after the Earth crashes into the Sun. "

ThanosBaskous:
Thanks @zekizeki!

zekizeki:
@QuinnyPig @ThanosBaskous I've created a new pull request here https://github.com/open-guides/og-aws/pull/438

PR #443, opened 2017-05-15
Gentle formatting fixes we can run more often.

jlevy:
Fixes a bunch of inadvertently preformatted text due to >4 spaces after a bullet.

The whole tab thing markdownfmt introduced is a pain but not doing a wholesale change just yet.


ThanosBaskous:
lgtm @jlevy, thanks.

PR #446, opened 2017-06-02
adding some important lambda limitations

mpetyx:
after some issues with those limitations, I thought it would be useful for more developers to understand and know upfront the current limits on lambdas


bgdnlp:
I'm not a native English speaker, might be wrong, but I"m pretty sure that "resource limits include" is correct, rather than "resource limits includes".

The other two limits are confusing to me. The way I read them is:
🔸 (...) deployment package size (.zip/.jar file) has a default limit of 50 MB (...)
🔸 (...) deployment package (uncompressed zip/jar size) can be at most 250 MB (...)

Does it mean 50MB packed, 250MB unpacked? Can it be rephrased to clarify a bit? Because at first glance it looks like both sentences talk about the same thing, deployment package size, but give different numbers.

mpetyx:
yes yes you are right. writing it as 50MB packed, 250MB unpacked would be much simpler. i will update it and create a new pull request

mpetyx:
@bgdnlp can you verify that it is better now?
Thanks for the input!

jlevy:
👍 Looks good; I've adjusted the language and formatting for clarity. Thanks @mpetyx and @bgdnlp !

PR #450, opened 2017-06-15
Fix indentation

hendi:
This fixes the indentation at two places, which resulted in incorrectly rendered markup.


jlevy:
👍 Thanks @hendi

Issue #455, opened 2017-06-15
TODO: Add Code of Conduct


QuinnyPig:
This one should have been done ages ago.

jlevy:
Fixed #456

PR #456, opened 2017-06-16
Create CODE_OF_CONDUCT.md

jlevy:
Add Contributor Covenant code of conduct.

Already discussed with @ThanosBaskous @QuinnyPig but we would appreciate any feedback (voice of support or concerns about this particular one, etc.) before we merge. DM on Slack if preferred.


ThanosBaskous:
Addresses #455

PR #458, opened 2017-06-24
add KMS gotcha about s3 multi part uploads

zoellner:
couldn't find any documentation about this. took me a few hours to get to the root cause.


bgdnlp:
Good one. Without checking, makes sense. You upload a bunch of parts, these are each encrypted. In order to complete the uploads S3 probably needs to decrypt all these parts, put them together, generate a new data key and encrypt the resulting file.

ThanosBaskous:
Merging after review by @bgdnlp and myself.

PR #460, opened 2017-06-30
Added bit about S3 race condition protection

QuinnyPig:
Unfortunately this was in a private conversation; can anyone find something to cite?


jlevy:
It's not clear to me which race conditions this is referring to — is it saying specifically that a Lambda function will not trigger on an S3 update and then see the out of date object in S3?

If we can't find a reference _and_ this but it seems to be known to be true let's add 📜.

jlevy:
@max-zanko or @ehammond may know something about this.

QuinnyPig:
That's exactly what I'm saying; I'm still digging out a reference on this. It comes down to the update PUT consistency model...

ehammond:
Since I was @-mentioned, I'll comment that I don't know what this is about.

bgdnlp:
In my opinion it sounds like the kind of thing one would only find relevant when it's very reliable. I mean, either you need it but it's not 100% reliable because it can fail during outages, or you accept that there's a chance of failure anyway, so it doesn't matter.

What I'm saying is, is it worth mentioning without it being documented

QuinnyPig:
I'm closing this, as I'm unable to validate this further. I suspect my source may well be wrong.

PR #468, opened 2017-07-12
Updated info regarding ALB Host header support

cloudowski:
Support for Host based routing was [announced](https://aws.amazon.com/about-aws/whats-new/2017/04/elastic-load-balancing-adds-support-for-host-based-routing-and-increased-rules-on-its-application-load-balancer/) some time ago.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1049,7 +1049,7 @@ ALB
 -	🔸ALBs only support HTTP/2 over HTTPS (no plain-text HTTP/2).
 -	🔸ALBs only support HTTP/2 to external clients and not to internal resources (instances/containers).
 -	ALBs support HTTP routing but not port-based TCP routing.
--	ALBs do not (yet) support routing based on HTTP “Host” header or HTTP verb.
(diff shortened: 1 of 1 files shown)


ThanosBaskous:
Thanks for catching this, @cloudowski!

I found the announcement for the addition of Host-based routing, so I'm in agreement that at least that part of the bullet should be changed. I haven't yet been able to find an announcement or documentation indicating that "HTTP verb"-based routing is available. If it is the case, can you link it?

I found this, which does not include any mention of HTTP verb:

> There are two types of rule conditions: host and path. When the conditions for a rule are met, then its action is taken.

http://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html#listener-rules

If HTTP verb-based routing is indeed still not possible, I propose that we remove the mention of it not being available (it's a relatively uncommon need, and I believe that I originally included it as an afterthought to the much more significant host-based routing gap).

ThanosBaskous:
Thanks again @cloudowski. I'm going to merge and then clean up the bullets describing routing capabilities a bit.

PR #481, opened 2017-08-14
Info about Red Hat Cloud Access for RHEL subscription portability to AWS

kprabhak:
Should save $$ on RHEL subscription costs in AWS if one already has previously paid for (and valid) on-premise RHEL subscriptions.


QuinnyPig:
Well spotted. Can confirm you're correct-- merging!

PR #488, opened 2017-08-19
S3 - adding note about consistency around the list operation.

neverendingqs:


QuinnyPig:
Well spotted, and you're absolutely right. Thanks for the contribution!

PR #490, opened 2017-08-21
add lambda failure case, /tmp storage limit.

gabinante:
Encountered a fun and confusing lambda failure case, and had to use my work account with enterprise support to figure out that it just meant my dynamodb stream was inactive not super broken like it appeared to be. 

Additionally, the 512mb /tmp storage limit is quite relevant for would-be lambda users (at least, it was for me)


QuinnyPig:
Nice! Is this a one-off, or is it easy to trip over? 

Don't want to fill the guide with corner cases, is all.  

The 512MB limit on /tmp is well spotted.

QuinnyPig:
Oh, and don't PR the authors.md file. That's autogenerated, which we should do again. Ahem. Paging @jlevy!

gabinante:
@QuinnyPig not a one-off. here's my message from AWS:

"The reason why you usually get this error is that when you set the lambda function to poll from Dynamo DB, it automatically keeps checking for latest data from Dynamo DB. However if there are no records for more than 48 hours, then lambda usually gives error "Internal Lambda Error".

To verify this, you can try entering new data into the DynamoDB table to see if the error disappears. 

Lambda adopts this mechanism because of the expensive pollers running behind. I would request you to please put some data to the Dynamo DB table and let us know if you are still facing the same issue. The lambda team is also aware of this issue and is working on making the error clearer in the future.

Hope this answers your query. Please let us know if you have any other questions."

QuinnyPig:
Holy crap. Yeah, let's include it. Just revert the change to authors.md and you're good to go.

Well spotted!

gabinante:
@QuinnyPig reverted authors

gabinante:
oh RIP, my aurora change is in here too which contributing tells me not to do. I was originally going to stage that as a separate PR; i can break it out if desired.

README.md
```diff
@@ -332,7 +332,7 @@ It’s important to know the maturity of each AWS product. Here is a mostly comp
 |------------------------------------------------------------------------------------------------------------|------------------|-------------------------------------------------------------------------------|:-----------:|:---------------:|:-----------------:|
 | 🐥[Lex](https://aws.amazon.com/releasenotes/Amazon-Lex?browse=1)                                           | 2016-11          | Preview                                                                       |             |                 |                   |
 | 🐥[Polly](https://aws.amazon.com/releasenotes/Amazon-Polly?browse=1)                                       | 2016-11          | General                                                                       |✓            |                 |                   |
-| 🐥[Rekognition](https://aws.amazon.com/releasenotes/Amazon-Rekognition?browse=1) 							 | 2016-11          | General                                                                       |✓            |					 |  				|
```
QuinnyPig: Nice trailing whitespace. :-p

QuinnyPig:
I'm happy with this; can someone else weigh in to confirm?

README.md
```diff
@@ -1349,11 +1350,12 @@ Lambda
...
 - 🔸 While adding/removing S3 buckets as triggers for Lambda function, this error may occur: "There was an error creating the trigger: Configuration is ambiguously defined. Cannot have overlapping suffixes in two rules if the prefixes are overlapping for the same event type." In this case, you can manually remove the Lambda event in the "Events" tab in the "Properties" section of the S3 bucket.
+- When using DynamoDB as a trigger for your Lambda functions, this error may occur: "PROBLEM: internal Lambda error. Please contact Lambda customer support." This usually just means that Lambda can't detect anything in the DynamoDB stream within the last 48 hours. If the issue persists, deleting and recreating your trigger may help.
 -	🔸 Lambda has several [**resource limits**](http://docs.aws.amazon.com/lambda/latest/dg/limits.html) as of 2017-06:
 	-	A **6MB** request or response payload size.
 	-	A **50 MB** limit on the compressed .zip/.jar file deployment package size.
 	-	A **250 MB** limit on the code/dependencies in the package before compression.
-
+	- A **512 MB** limit on local storage in /tmp
```
bgdnlp: Nitpick, but maybe add a dot at the end, just to be consistent with the other points?

Also, should we add on the same line that /tmp is the only place where files can be written to on a Lamda container?

None of these are blockers, they can be added later.

QuinnyPig:
Looks good to me. Thanks!

PR #494, opened 2017-08-28
Updated Amazon Webservices for Dummy link

victorelu:
It was returning 404 on the dummies.com website, but the amazon.com link should be permanent.


bgdnlp:
Trying to access their store form their own site, http://www.dummies.com/programming/cloud-computing/amazon-web-services/amazon-web-services-for-dummies/, also returns a 404, it points to the same link.

I'd go with the Amazon link.

QuinnyPig:
Confirmed. Good catch-- thanks!

PR #502, opened 2017-09-09
Added more details to AWS Batch Section

lobster1234:
Added more details based on the feedback received on slack.


jlevy:
👍 Looks good. One request would be if we can add a few links to some of these items — ideally most of the bullets would have a link.

I'm good with it as is, but it should get a 2nd review from someone with more Batch expertise.

jlevy:
And a nit: docker -> Docker throughout.

lobster1234:
@jlevy Thanks for the feedback! I will add a few links and fix the docker vs. Docker 👍

ThanosBaskous:
Thanks for the edits, @lobster1234!

PR #516, opened 2017-09-30
Added region specificity gotcha for KMS

lobster1234:
This is something that can be overlooked unless the key ARNs are observed, but most developers only deal with key IDs. The FAQs for KMS do cover it.


lobster1234:
@ThanosBaskous Thanks so much!

PR #519, opened 2017-10-05
First stab at addressing per-second billing

QuinnyPig:


bgdnlp:
* The Scalr link returns 404 for me. Is this the new way: https://github.com/Scalr/installer-ng ?

* Would it be worth mentioning that hourly billing only applies to dedicated instances now? Assuming that that's true.

QuinnyPig:
Hah, good find. I didn't touch the link yet. Will fix while I'm here.

QuinnyPig:
> Would it be worth mentioning that hourly billing only applies to dedicated instances now? Assuming that that's true.

Nope-- it also applies to Windows instances, and per-hour licensed enterprise Linux instances. 

I suggest merging this one, and I'll come up with a separate section on per-second billing this week.

bgdnlp:
FWIW, agree with the merge.

ThanosBaskous:
Good cleanup @QuinnyPig. Thanks for the review @bgdnlp. I agree with a merge on this + separate section to follow.

PR #523, opened 2017-10-20
Added the SSE-SQS <-> SNS gotcha

lobster1234:
I wrote a blog demo-ing this gotcha - https://lobster1234.github.io/2017/10/14/fan-out-with-sns-and-sqs-gotcha/.

Could not find it documented anywhere in the official docs per my extensive search, but this one bit us hard during a production rollout as we use SSE SQS in prod but not in lower environments.


QuinnyPig:
This is awesome-- thanks!

ThanosBaskous:
👍 Thanks @lobster1234!

PR #524, opened 2017-10-21
Scalr is no longer open source apparently.

QuinnyPig:


QuinnyPig:
This is the second time in a month the Scalr link has come up as a 404. I'm tired of chasing that down; if there's another good open source example we can include it instead.

ThanosBaskous:
👍 Thanks @QuinnyPig, I agree.

PR #527, opened 2017-10-22
Update ownership information and link for Ice.

QuinnyPig:


ThanosBaskous:
Thanks @QuinnyPig.

For future reference, the takeover of Ice by Teevity from Netflix can be confirmed here:

https://twitter.com/NetflixOSS/status/870475268166606851

PR #531, opened 2017-10-24
Add "stacker" tool to CloudFormation Tips

ThanosBaskous:
cc @phobologic


phobologic:
How I didn't do this before, I have no clue.  LGTM, thanks @ThanosBaskous .  Do you tend to link to github repos, or docs, because stacker has full docs at http://stacker.readthedocs.io/en/latest/ as well.

README.md
```diff
@@ -1489,6 +1489,7 @@ CloudFormation
 	- Troposphere does not support all of the resources types you can describe with CloudFormation templates.
```
phobologic: I think this is a false assumption due to poor docs in the currently supported section.  That should probably just be removed, it changes so often.  We're usually pretty close to up to date.

ThanosBaskous: 👍 Updated to reflect this.

ThanosBaskous:
👍 @phobologic updated to point to docs instead of repo.

phobologic:
LGTM!

PR #532, opened 2017-10-26
Whitelisting hacker news for firewall issues in link checker

QuinnyPig:


ThanosBaskous:
👍 lgtm

.travis.yml
```diff
@@ -1,4 +1,4 @@
 language: ruby
 rvm: 2.2
 before_script: gem install awesome_bot
-script: awesome_bot README.md --allow-redirect --allow-dupe --allow 429 -w .amazonaws.com
\ No newline at end of file
+script: awesome_bot README.md --allow-redirect --allow-dupe --allow 429 -w .amazonaws.com,news.ycombinator.com
```
baynezy: @QuinnyPig - it would have probably been better to `--allow 429,520`

QuinnyPig: Doesn't that open the door for other sites to throw 520s? We know Hacker News is going to block it, but no other sites should...

baynezy: That is true, but now you will never detect a 404 from Hacker News. This script is to detect 404s specifically, whereas a 520 is usually a temporary issue. Ultimately it is your call, but I feel it is a bit more nuanced.

QuinnyPig: If it helps any, I have this script set to run without the HackerNews whitelisting as a git hook on my side; Hacker News errors will be caught eventually!

baynezy: Fair enough. As I said it is your call.

Issue #308, opened 2016-11-10
Mention S3 server-side encryption


jlevy:
https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html

Deserves a basic explanation and a link also to (no expanding) KMS section.

ajkerrigan-mdsol:
This is a related discussion from #questions today:

ahmed [9:11 AM] 
I think I already know the answer to this question but need confirmation. With SSE-S3 since Amazon is managing the encryption technically if I’m using SSE-S3 for both buckets I can decrypt the contents of both with a regular IAM user that has access to the buckets since it’s just AES256, right?

jose [9:14 AM] 
Yup should be transparent, as long as the user has permission to the bucket S3 will handle the decryption.

ahmed [9:15 AM] 
Thought so. If I want to manage the decryption access then I need SSE-KMS

jose [9:16 AM] 
Well not really, you'll want Client-Side encryption for that SSE-KMS is just telling S3 which keys to use to encrypt it, but access is the same since it's still server side

[9:17]  
Really this is just in case someone steals an entire data center worth of drives and your info is on it

[9:18]  
Client Side encryption would store the file as unreadable to even S3, the client would encode and decode

ahmed [9:19 AM] 
Okay hang on a sec. CloudTrail uses SSE-KMS though and you need to grant decryptKey to the IAM user to access the objects.
http://docs.aws.amazon.com/kms/latest/developerguide/services-cloudtrail.html
http://docs.aws.amazon.com/awscloudtrail/latest/userguide/create-kms-key-policy-for-cloudtrail-decrypt.html
docs.aws.amazon.com
How AWS CloudTrail Uses AWS KMS - AWS Key Management Service
Learn how AWS CloudTrail uses AWS KMS to encrypt log files. 
 docs.aws.amazon.com
Granting decrypt permissions - AWS CloudTrail
Grant other users permissions to decrypt logs with a CMK. 

[9:19]  
So CloudTrail is doing object encryption on top of SSE-KMS? (edited)

jose [9:21 AM] 
CMK

[9:21]  
http://docs.aws.amazon.com/AmazonS3/latest/dev/serv-side-encryption.html
docs.aws.amazon.com
Protecting Data Using Server-Side Encryption - Amazon Simple Storage Service
How to protect data using Server-Side Encryption in Amazon S3. 

[9:21]  
3 Options

ahmed [9:21 AM] 
Right

[9:22]  
[...]

PR #534, opened 2017-10-29
Add special topic on encryption

bgdnlp:
The PR is intended as a first draft, request for comments. Is it useful? Is it in the right place, should it be formatted differently?

The encryption it presents relates mostly to S3, but not only. The same process applies to other services, which is why I didn't want to link it to S3 specifically.

Is it useful?

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Table of Contents
 
 -	[High Availability](#high-availability)
 -	[Billing and Cost Management](#billing-and-cost-management)
+-   [Encryption](#encryption)
 -	[Further Reading](#further-reading)
 
 **Legal**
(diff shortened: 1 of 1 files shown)


Ashex:
There's some good info here too: #308 

Biggest caveat for S3 is that cross region replication is currently only supported with SSE-S3.

bgdnlp:
That's one of the questions, should this section be included under S3, or should it be left separate as general encryption topic and maybe referenced from S3?

Would be good to add examples for bucket policies enforcing SSE-KMS or SSE-S3, but that doesn't make sense if it's a general topic, so I left them out.

Ashex:
Considering you're focusing on S3 encryption it would probably make more sense for it to fall under S3. 
Considering encryption in the context of AWS is usually going to be KMS perhaps the section on envelope encryption should then be moved to go under KMS?

The KMS area needs some extra love that I've been meaning to expand on so I can fill in some info on key policies and CMKs if you can create a quip document that I can contribute to.

bgdnlp:
Turned this pull request into a Quip document. "Encryption Types" in the og-aws folder.

Should this PR be closed then?

QuinnyPig:
Closing this PR. Also, let’s make sure that we capture the new S3 encryption stuff!

PR #536, opened 2017-11-01
Add tips and gotchas to API Gateway

andre-meireles:


andre-meireles:
Gonna redo this without the ugly merge commits

PR #538, opened 2017-11-01
Add tips and gotchas to API Gateway

andre-meireles:


QuinnyPig:
This looks good to me. One more person should take a look at it prior to merging.

ThanosBaskous:
lgtm! Thanks @andre-meireles!

cristim:
I would like to see a limitation/gotcha about the lack of multiregion support: when using custom domains you can only have a single API gateway per domain so it is impossible to get multiple endpoints configured in a  multiregion redundant setup for the same domain.

PR #545, opened 2017-11-25
Updates for recent announcements

QuinnyPig:


QuinnyPig:
There are. This is just what jumped out at me as I was on a plane earlier tody.

jlevy:
👍 LGTM and verified the SLA change number.

PR #549, opened 2017-11-28
Updated SQL Server storage limitations

chuckbelcher:
Updated the SQL Server storage limitations to reflect the increase from 4TB to 32TB


chuckbelcher:
Initial pull request had an incorrect update. MSSQL now supports 16TB, not 32TB. A second commit was added to correct the error.

jlevy:
Thanks @chuckbelcher ! Do you have an updated link we can cite with those numbers?

chuckbelcher:
The updated storage limits can be found here http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Storage.html

PR #551, opened 2017-11-30
Add note about ALB/CLB/NLB limit increases

matthiasr:
based on a real life experience trying to raise all 3 limits at once.


ThanosBaskous:
Thanks @matthiasr - can you speak to what the default limits are?

QuinnyPig:
ALBs and CLBs share a default limit of 20, and NLBs get a separate 20. Both are per region.

matthiasr:
Are you sure about the shared ALB/CLB limit? For me they are listed separately, and the limit increase request form distinguished between them.

PR #555, opened 2017-12-04
ipv6 has been worldwide since January

QuinnyPig:


ThanosBaskous:
👍 Thanks @QuinnyPig and @cristim!

PR #556, opened 2017-12-04
Added wording about DNS validation for certificates

QuinnyPig:


ThanosBaskous:
👍 Thanks @QuinnyPig and @cristim!

PR #557, opened 2017-12-04
Updated hypervisor info for Nitro and bare metal

QuinnyPig:


ThanosBaskous:
Nice. Thanks @QuinnyPig!

PR #561, opened 2017-12-10
Added new updates to CloudWatch service.

lobster1234:
* Alerting on M out of N metric against threshold
* KMS Encryption of CloudWatch Logs


QuinnyPig:
Correct on both accounts. Merging. Thanks for the help!

PR #562, opened 2017-12-10
Clarify and emphasize S3 bucket vs object permissions

bgdnlp:
Hope I didn't miss it, but while the guide does mention that making a bucket private is not enough, it's not very clear.


QuinnyPig:
This is awesome stuff. Can we get one more person to weigh in?

PR #575, opened 2017-12-30
added aws-lambda-list to Lambda Code Samples

darkuncle:
h/t #lastweekinaws


QuinnyPig:
Nice work. Thanks for pitching in!

PR #576, opened 2018-01-02
Load Balancer Tip on retaining IAM Server Certificate ARN

Ashex:
Once upon a time there was a page in the AWS Documentation describing how to rotate IAM Server Certificates without having to do a total replacement on the load balancer.

I did some research and sorted the process for doing this and have added a high level writeup to the load balancer tips on how to achieve this. This may not be the appropriate place to add it since this is technically IAM but since it is specifically for CLB/ALB felt it was the most suitable area.


QuinnyPig:
I remember that page too-- and I can't find it either. Ugh. Thanks for pitching in.

PR #579, opened 2018-01-06
Fix a couple typos

ryanguest:


QuinnyPig:
Glorious! Thanks for the contribution-- and you should be very pleased with your sharp eyes!

PR #581, opened 2018-01-13
Add additional gotchas to CloudFormation

andre-meireles:
Just a couple of cases I have run into lately.


cristim:
While at it,speaking of stack deletion issues, maybe we should mention that sometimes stacks fail to be deleted if their resources are still in use, and how deletion can be forced by ignoring those resources. Also in case of custom resources there is a workaround by making a HTTP request to their result URL.

ThanosBaskous:
Great gotchas, @andre-meireles! Are those (default) API limits documented anywhere that could be linked as a part of this PR?

QuinnyPig:
Thanks for the PR; added the citation myself, merging now!

ThanosBaskous:
👍  Thanks everyone!

PR #582, opened 2018-01-18
Limitations of using Lambda with VPC resources.

thomasmichaelwallace:
Mentions why you might need to run Lambdas within a VPC and the main challenges:
* Slow cold starts.
* Relatively low default limit of ENIs, which AWS Support may be reluctant to immediately raise.

(This was prompted by an experience and resulting discussion in the AWS channel of the Serverless forum slack)


QuinnyPig:
Well spotted. Thanks for helping with the project!

PR #585, opened 2018-01-27
Add note on API Gateway query params

shtaft:
API gateway does not have full support for nested query parameters. Based on personal experience and their [docs](https://docs.aws.amazon.com/apigateway/api-reference/making-http-requests/ ). See the *HTTP request query parameters* section


QuinnyPig:
Much to our everlasting frustration, you are correct. Thanks for pitching in!

PR #590, opened 2018-02-22
Added Curated List of Resources

qualityjacks:
https://hackr.io/tutorials/learn-amazon-web-services-aws


jlevy:
@qualityjacks thanks for the addition.

It looks like the list is mostly basic items with 0 upvotes however. Am I missing something?
image

If it's just that set of links I don't think it merits inclusion as is.

QuinnyPig:
While the content is fairly solid, the way that Hackr.io presents it doesn't rise to the quality bar that merits inclusion in the guide at this time. Thanks for the contribution, though!

PR #594, opened 2018-03-03
Add awsgeek to list of resources

pgrzesik:
I'm a big fan of Jerry Hargrove's blog (his hand-drawn diagrams are awesome!) so I though that it might be a good idea to add this resource to open guides.


QuinnyPig:
Thanks-- looks great!

PR #598, opened 2018-04-03
Added general reference: AWS cloud cost checklist

mobrien:
Added to general references a checklist document to help with lowering AWS cloud costs.

[Lower AWS Cloud Costs Checklist](https://www.powerdown.io/blog/posts/stories/checklist-to-lower-aws-cloud-costs.html)

Disclosure: I authored the checklist and work for PowerDown that is briefly mentioned in the paper.


QuinnyPig:
Thanks for the contribution; do you have any affiliation with that site?

cristim:
@QuinnyPig: the "Disclosure" note states that @mobrien is the author of that checklist

QuinnyPig:
...I can't read. Thanks. :-)

jlevy:
@mobrien thanks! While the post has good info in it, it's brief and not definitive or well known, like everything else we list in this section. So I don't think this is the right place for that post.

It'd be great to have a few of the specific tips included though! Your blog has good cost tips. Would you be up for actually including those suggestions instead in the [cost section](https://github.com/open-guides/og-aws#billing-and-cost-management)?

In general, we're very glad to include individual tips and if they are complex, link to supporting blog posts that explain them in more detail.

PR #602, opened 2018-04-30
Update AWS Fargate regions

manelpb:
https://aws.amazon.com/about-aws/whats-new/2018/04/aws-fargate-now-available-in-ohio--oregon--and-ireland-regions/


QuinnyPig:
Thanks for the contribution-- you're on the ball!

PR #606, opened 2018-05-30
Adding some KMS Gotchas 

salexpdx:
Today's lesson tracking down why I can't view/manage/use a KMS key seems worth sharing


bgdnlp:
So, not even the root account can access that key? Only support can solve it?

salexpdx:
That's correct.  I added a link to a doc that mentions the risk of keys becoming unmanageable.

QuinnyPig:
Awesome-- thanks!

PR #610, opened 2018-06-19
Remove wrong statement about security groups

marco-hoyer:
Security groups are bound to a specific VPC and thus to an AWS region but not to a single AZ or DC. I would like to remove this statement to prevent confusions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1643,7 +1643,6 @@ VPCs, Network Security, and Security Groups
 
 ### VPC and Network Security Gotchas and Limitations
 
--	🔸Security groups are not shared across data centers, so if you have infrastructure in multiple data centers, you should make sure your configuration/deployment tools take that into account.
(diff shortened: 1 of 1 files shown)


DavidTPate:
Yeah that sentence definitely doesn't make much sense. 

I think the overall point is correct it is just using the incorrect terms. I'd rather see it reworded to something like this:

> Security groups are not shared across VPCs or Accounts, so if you are utilizing infrastructure in multiple VPCs or Accounts you should make sure your configuration/deployment tools take that into account.

From my look through the VPC/Network section I don't see a mention of the above.

networkpadawan:
please be mindfull that security groups for vpc peering across accounts exist.
https://aws.amazon.com/about-aws/whats-new/2016/03/announcing-support-for-security-group-references-in-a-peered-vpc/

marco-hoyer:
@networkpadawan you're right but I would argue it is a bit different. Security groups are still bound to one VPC/Account/Region but they can be referenced in another VPC/Account/Region when VPC peering is applied.

marco-hoyer:
@DavidTPate like your approach. I extended it a bit. Better?

QuinnyPig:
Thanks for the hard work!

PR #612, opened 2018-07-13
Removing gotcha from sqs regarding lambda

apexkid:
SQS can now be used as a lambda trigger.


QuinnyPig:
Perfect! Thanks for the contribution!

PR #611, opened 2018-07-10
adding NLB gotcha on client connectivity

DaleBetts:

PR #614, opened 2018-07-17
Fix the link for the gotcha to take us to the correct page

DavidTPate:
Updates the link added in #611 so that we point to the relevant section of the documentation for the gotcha as opposed to just the main page.


QuinnyPig:
Great find. Thank you! Feel free to join us in the Slack workspace if you haven't already...

Issue #615, opened 2018-07-20
Broken link


edgarpf:
Link in "include versions or hashes in filenames" leads to 404 error.

edgarpf:
salexpdx solved the issue.

DavidTPate:
Please close the issue @edgarpf if it is resolved.

PR #616, opened 2018-07-21
Fixing Issue #615 Broken link on CloudFront url.  URL structure chang…

salexpdx:
…ed slightly on domain and it no longer uses /post before the post name.


salexpdx:
Broke the ideas into two separate bullet points.  I left out the comment about T2, I feel like I should have an article backing the statement to explain it but didn't find much on a google search.

README.md
```diff
@@ -855,6 +855,11 @@ EC2
...
+- AWS [introduced Dedicated Tenancy](https://aws.amazon.com/blogs/aws/amazon-ec2-dedicated-instances/) in 2011. This allows customers to have all resources from a single server. Some saw this as a way to solve the [noisy neighbor problem] (https://www.infoworld.com/article/3008225/cloud-computing/amazon-dedicated-hosts-bye-bye-to-noisy-cloud-neighbors.html) since only that customer uses the CPU. This approach comes with a significant risk if that physical system needed any type of maintenance. If a customer had 20 instances running using shared tenancy and one underlying server needed maintenance, only the instance on that server would go offline. If that customer had 20 instances running using dedicated tenancy, when the underlying server needs maintenance, all 20 instances would go offline. 
+
```
QuinnyPig: May also want to include the tip that significantly large instances get their own hardware regardless of tenancy selection. And then get rid of the extra lines here...

PR #617, opened 2018-07-24
Added Route 53 Gotcha

FoxUSA:
Added Route 53 Gotcha

Sources:
> Route 53 will only respond to queries for these names when the queries originate from within the VPC(s) that you authorize. 

https://aws.amazon.com/about-aws/whats-new/2014/11/05/amazon-route-53-now-supports-private-dns-with-amazon-vpc/

> If you need to perform name resolution of both your VPC and on-premises workloads, additional configuration is required.

> If you have integrated your on-premises network with one or more Amazon VPC virtual networks and you want your on-premises network to resolve domain names in private hosted zones, you can create a Simple AD directory. Simple AD provides IP addresses that you can use to submit DNS queries from your on-premises network to your private hosted zone.

https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/hosted-zone-private-considerations.html

>  Customers with workloads that
leverage both the AWS VPC and on-premises resources often need to resolve
private DNS records hosted on-premises too. Currently, this can’t be
accomplished by the Amazon provided DNS server alone. 

https://d1.awsstatic.com/whitepapers/hybrid-cloud-dns-options-for-vpc.pdf


alexandernst:
Can we merge this?

QuinnyPig:
Thanks for the contribution-- sorry for the delay!

PR #618, opened 2018-07-24
Remove quote about sns lambda vs sqs not lambda

M1ke:
Helpfully this is no longer true!


QuinnyPig:
Correct! Merging. Thanks for the contribution!

PR #619, opened 2018-07-24
Alphabetise top list

M1ke:
Easier to read the top list and find a specific service

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,50 +25,50 @@ Table of Contents
 
 | Specific AWS Services                 | Basics                         | Tips                          | Gotchas                                        |
 |---------------------------------------|--------------------------------|-------------------------------|------------------------------------------------|
-| [Security and IAM](#security-and-iam) | [📗](#security-and-iam-basics) | [📘](#security-and-iam-tips) | [📙](#security-and-iam-gotchas-and-limitations) |
-| [S3](#s3) | [📗](#s3-basics) | [📘](#s3-tips) | [📙](#s3-gotchas-and-limitations) |
-| [EC2](#ec2) | [📗](#ec2-basics) | [📘](#ec2-tips) | [📙](#ec2-gotchas-and-limitations) |
-| [CloudWatch](#cloudwatch) | [📗](#cloudwatch-basics) | [📘](#cloudwatch-tips) | [📙](#cloudwatch-gotchas-and-limitations) |
-| [AMIs](#amis) | [📗](#ami-basics) | [📘](#ami-tips) | [📙](#ami-gotchas-and-limitations) |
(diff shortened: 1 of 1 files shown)


QuinnyPig:
I like this a lot, but it does change the structure of the guide somewhat. Would appreciate others' feedback prior to merge.

scott-cbt:
This seems like one of those changes that is so obvious after the fact, but something that people who are used to the project wouldn't think of.  I think it will make it easier for new people to find the section they want since they won't have to use find in document and can just scroll the list in order. 

As for service area vs alpha I think that adding more consistency might help with that.  Right now there is CLB, ALB, and Load Balancers as topics.  If they were all EC2 Load Balancers - Elastic [ELB], etc that would keep the three together, and make the alpha order very easy.

salexpdx:
This seems like one of those changes that is so obvious after the fact, but something that people who are used to the project wouldn't think of. I think it will make it easier for new people to find the section they want since they won't have to use find in document and can just scroll the list in order.

As for service area vs alpha I think that adding more consistency might help with that. Right now there is CLB, ALB, and Load Balancers as topics. If they were all EC2 Load Balancers - Elastic [ELB], etc that would keep the three together, and make the alpha order very easy.

M1ke:
@QuinnyPig @salexpdx I'm happy to do that change too if it will help. I'd suggest the same with Glacier; you can't use Glacier without s3, there's no separate interface for it, so S3 - Glacier would make more sense in the ordering, other than that some people might have heard of Glacier as a brand.

I also wonder if SES, SQS & SNS are well enough known acronyms and could be improved by noting the name - then someone who's starting to read with the question "does AWS have a message queue" gets their question answered reading the index. I believe S3 and EC2 are big enough brand names they don't need to be explained in the same way.

M1ke:
@QuinnyPig @salexpdx any update on this? The Travis build failed but looks like it's due to something unrelated, a URL check that failed but I've not modified any URLs

QuinnyPig:
Yeah, ignore the Travis results. I think you two are correct; if you get this to merge cleanly I'll endorse it!

M1ke:
@QuinnyPig Have now resolved the conflict so we can merge this

QuinnyPig:
It would seem not. One more conflict resolve, and I'll merge it!

M1ke:
Ok I'll check it tomorrow, it was merging cleanly when I last commented but
guess other stuff has changed

maishsk:
@M1ke we still have some conflicts - can you have a look please?

maishsk:
@M1ke we still have some conflicts - can you have a look please?

QuinnyPig:
Glorious! Merging.

maishsk:
LGTM

PR #621, opened 2018-07-29
Update README.md

AstroTom:
Bug with spaces at the end of Tag keys

There is a bug in the ASG console where spaces after tag names are preserved. So if you type "Name " with an extra space at the end you will not get the expected behavior. This is probably true in other locations and SDKs also. Be sure you do not add trailing spaces to tag keys unless you really mean it.


QuinnyPig:
Holy crap. Is this still the case? If so it's absolutely worth opening a feedback case with AWS! Merging in for the time being.

wishtrip-dev:
Yes, I have opened a case. But as you know AWS always says they can provide any ETA for a fix.
Thanks for the merge!

AstroTom:
Yes, I have opened a case. But as you know AWS always says they can provide any ETA for a fix.
Thanks for the merge!

PR #628, opened 2018-09-07
Added PostgreSQL

vventirozos:
Added PostgreSQL under RDS separated RDS Aurora into RDS aurora postgres and RDS aurora MySQL.
also fixed the ordering for RDS SQL server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1221,21 +1221,21 @@ RDS PostgreSQL
 
 ### RDS PostgreSQL Basics
 
-- RDS offers PostgreSQL 9.3, 9.4, 9.5, 9.6, and 10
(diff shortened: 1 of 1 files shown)


xzilla:
FWIW, I worked with @vventirozos to put together the above changes. We work at OmniTI which is an RDS partner and Major Sponsor of the Postgres project, so we feel pretty good about the accuracy of the information. If anyone has questions they can ping me on the og-slack. 

Side note, it looks like the build failures are for failing urls unrelated to our changes, so hoping we can get this merged without having to fix those.

QuinnyPig:
Very well done. Thanks for the contribution!

PR #629, opened 2018-09-07
Update outdated GPU section

lukedeo:
AWS's GPU offering have changed quite a bit since this was last updated. In addition, the decision space around deep learning AMIs is now a fair amount more nuanced.


QuinnyPig:
Thanks— great catch!

PR #630, opened 2018-09-08
Presentation

kaivan-s:
Presentation of all the services provided by AWS with brief information.


QuinnyPig:
Thanks for the contribution! I'm relatively unconvinced that including a PPT in the repository is the best way to present the information-- it'll significantly bloat the repository size, for one. Is there new information contained in this that isn't available elsewhere? If so, uploading to SlideShare and linking against it may be the much better approach here.

PR #633, opened 2018-09-28
Add gotcha about S3 bucket deletion

dtjm:


QuinnyPig:
Thanks-- great addition!

PR #637, opened 2018-10-07
Added openstack components

ramlaxman:
- Added Openstack, private IaaS cloud.
- As most of its services inspired from AWS, so mentioned same ones here.


ThanosBaskous:
I'm fine with adding it. Do you have a sense for how much momentum / use it sees these days? I'm a little bit out of the loop on OpenStack.

Also, this can be corrected post-merge, but OpenStack is incorrectly capitalized in this MR.

maishsk:
OpenStack is slowly but surely becoming irrelevant - but this is a whole other theological discussion

PR #638, opened 2018-10-09
Fix noisy neighbor link markdown syntax

juaoose:
There was an extra space that broke syntax in the 'EC2 Gotchas and Limitations' section.


maishsk:
LGTM 👍

PR #639, opened 2018-10-15
Adds detail about cross-region RDS Aurora read replica limit

NeilsC:
This simply adds a note, from a recent AWS blog entry, on the limit of cross-region read replicas for an Aurora cluster.


maishsk:
👍  LGTM

PR #647, opened 2018-10-21
Added tip for WAF logging capability

pachulo:
Since the end of August is possible to get full logs of all the web requests inspected by the WAF.


maishsk:
Thanks @pachulo for submitting this,

LGTM :+1:

PR #649, opened 2018-10-23
Added caveat about unpublished service limits

QuinnyPig:


maishsk:
@QuinnyPig Do you think it would be useful to give an example?

QuinnyPig:
If you've got a good one that isn't NDAd, absolutely!

maishsk:
Point well made :) 

LGTM 👍

PR #656, opened 2018-10-28
Add info about possibility of running Android emulators on EC2 instances

artemnikitin:


QuinnyPig:
Is "running Android emulators" enough of a common use case for these instances, or is it a weird edge case?

artemnikitin:
I would say that this is a weird case. But if you want to do it, then you can do it only on **i3.metal**. All other instance types doesn't have access to some hardware virtualization features which are needed for that.

Issue #546, opened 2017-11-27
Add Step Functions


dkavanagh:
I want to confirm I didn't miss something. I recently gave a talk on Step Functions after using them for a project. I'd be happy to kick start this if it's actually missing.

jlevy:
Yes, indeed it is not a service we yet cover at all.

Would be great if you'd like to get a first draft out! (Want to link your talk here?) Best way is just to start with a Quip doc that has similar structure to an existing section and then discuss/improve in Slack.
See https://github.com/open-guides/og-aws/blob/master/CONTRIBUTING.md#please-help and def join us in Slack! @QuinnyPig @ThanosBaskous and I or other editors can give you feedback as you go.

dkavanagh:
Awesome, will do.

omenking:
I had thought maybe it already existed only to find the links are dead.
Is it common to pre-link before seeing an article? I ask since then I have an idea what is missing and what is not.

![screen shot 2018-08-19 at 5 43 47 am](https://user-images.githubusercontent.com/7776/44307595-e053fd00-a372-11e8-8819-c543ffb9e4c1.png)

maishsk:
Thanks @gliptak

PR #657, opened 2018-11-03
Correct Step Functions links

gliptak:
https://github.com/open-guides/og-aws/issues/546


gliptak:
@maishsk please review

PR #681, opened 2019-01-22
Add Fargate task storage gotcha

ScottBrenner:
Adding [Task storage is ephemeral. After a Fargate task stops, the storage is deleted. ](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-task-storage.html) as a Fargate gotcha.


ScottBrenner:
Merge?

PR #693, opened 2019-03-08
Add info about Step Function execution limits

hayfield:
This limit is not clearly documented and reduces the range of situations that it's appropriate to use Step Functions for iteration.


maishsk:
Thanks!!

PR #695, opened 2019-03-11
Update EKS Autoscaling Limitation

dkeightley:
Cluster-autoscaler is designed for this purpose.


vbvkel:
Agreed, this is the more proper way of doing that.

PR #706, opened 2019-06-06
Update VPC section on data transfer cost savings with VPC Endpoints

chrissng:
Update VPC section on data transfer cost savings with VPC Endpoints. Added a tip on using Gateway VPC endpoint to route data transfers between EC2 in private networks and S3 to save costs.


QuinnyPig:
Sorry this got dropped. Great contribution--thanks! Merging.

PR #707, opened 2019-06-23
Fix bold hyperlink for EC2 pricing

juaoose:
It is probable that the leading emoji was causing the rendering issue.

*DNS marked share link probably timed out during travis build.*


QuinnyPig:
Thanks--good find!

PR #716, opened 2019-09-09
Update README.md: API Gateway limitations

volkanp-ieu:
Limitations to what resources API Gateway can reach within a VPC.

For instance, you are unable to do basic proxying from a regional, edge-optimized or private API Gateway endpoint to a web service hosted within a VPC without an NLB deployed and placed upstream of that service. There are several problems with this, namely, you are forced to use an NLB, instead of an ELB or an ALB which are L7 load-balancers.

You also cannot proxy to a web service hosted on an EC2 instance that is in a private subnet. The EC2 instance has to be publicly accessible.


QuinnyPig:
Thanks for the contribution! While I agree that this isn't ideal, "it won't scale" doesn't seem accurate. It will, it just gets expensive unless I'm missing something?

volkanp-ieu:
You're right, scaling isn't the issue here, the fact that you are forced to use a layer-4 load-balancer as a reverse proxy introduces problems. Take for instance, an NLB that is upstream to an ALB which is a layer-7 load-balancer with all the path based routing and SSL offloading features, the problems arise when the ALB IP addresses change, you are quickly left with a broken API Gateway integration as the listeners are IP based.

I'm happy to amend the comment to reflect the above.

QuinnyPig:
Excellent; I'll merge it in once you do. Thanks for the contribution!

QuinnyPig:
Is this still accurate in light of https://aws.amazon.com/about-aws/whats-new/2019/09/amazon-api-gateway-simplifies-invoking-private-apis/ ?

PR #717, opened 2019-09-19
Certification updates

paulwakeford:
Adding ML and Alexa specialty certs.
Adding a line on training for certs, and on cert lounges.
Note: I work with (but not for) A Cloud Guru.


QuinnyPig:
Nice work. Rewording title for posterity.

PR #722, opened 2019-10-10
Reword how ACM domain validation works

QuinnyPig:
Domain validation is the superior / recommended method; putting it first instead of as an afterthought.


paulwakeford:
The way this is worked sounds like DNS validation is only supported with R53. Perhaps reword to: 'If the zone lives within Route 53 and the user has access, the necessary record can be added in the console via a single click during the certificate request process. If the zone is not within Route53 the user is required to update DNS manually. This is still preferred to the second way, which requires more user interaction...'

QuinnyPig:
This was updated. Thanks, @paulwakeford. Good to merge?

paulwakeford:
Minor typo :'If the zone is not within Route5 3'

QuinnyPig:
There I fixed it

QuinnyPig:
@paulwakeford Happy for me to merge this?

paulwakeford:
Ecstatic!

PR #724, opened 2019-10-10
Removed a dead link that is causing CI to fail

maishsk:
This dead link is causing CI to fail. The domain has expired.


QuinnyPig:
Nice find. Merging.

PR #726, opened 2019-10-17
Update link to ElastiCache User Guide

akamch:
Replaces broken link with two URLs to separate user guides for Redis and
Memcached.


QuinnyPig:
Good catch; merging. Thank you!

PR #731, opened 2019-10-31
Add Linode to "Other Providers" Column 

AKluger:
I noticed Digital Ocean was the only provider listed in this column in a number of rows so I added Linode where they have a comparable product:  Compute | Block Storage | Object Storage


QuinnyPig:
We generally don't accept PRs about other providers from representatives of those providers; down that path lies "this project becoming an advertising hot zone."

AKluger:
I get this. But for the record, I don't work in marketing and no one at the company told me to do this, I found your guide when trying to learn more about how we stacked up against AWS and thought this would help people.

PR #732, opened 2019-11-27
Update Slack invite links

jlevy:
Let's try using Slack's direct method.

These do have to be renewed every 2K invites, we believe, but they are less brittle, as we have run into issues keeping the Slack inviter up reliably (both infra hosting it and the fact that Slack seems to throttle).


sneko:
Hi @jlevy @maishsk , it seems the link has expired (https://join.slack.com/t/og-aws/shared_invite/enQtODM2NjY1NDQ2MTgxLWYwY2VjZDBiOGU1YTJjOWUwNTY3NjEyODA0NzY1N2MxNjhhZmYwZTU0NjNhMjNlNGVjODdlNTI4N2Y1YWIwNGE).

Could you share the new one please?

paulwakeford:
Use slackhatesthe.cloud.

Issue #720, opened 2019-10-03
the provided Slack channel (og-aws) Doesn't work


alrah003:
Hello,

I'm interested in joining your group on slack but I couldn't log in via the link provided on the GitHub. can you enroll me in the group?

FernandoMiguel:
@alrah003 you mean https://og-aws-slack.lexikon.io/ ?

alrah003:
yes, that what I meant.

alrah003:
I've already signed up but didn't receive the invite on my inbox

IsmaelMartinez:
same here.

Helicer:
Hello - encountering this issue too. I've gotten email bounces while writing to the email address also. Curious if anyone was able to get in / invite? Thanks!

hopkinsnji:
What's the status in this? I need to join!

prakashkumar007:
Same here,the invite is not working and the workaround to.

fotag:
same here

PaulChristophel:
Having this issue as well

aharper273:
This is the link I found hyperlinked in the og-aws github and it seems to work.

https://og-aws.slack.com/join/shared_invite/enQtODM2NjY1NDQ2MTgxLWYwY2VjZDBiOGU1YTJjOWUwNTY3NjEyODA0NzY1N2MxNjhhZmYwZTU0NjNhMjNlNGVjODdlNTI4N2Y1YWIwNGE

IsmaelMartinez:
Added a wee PR to add that as an extra link. Thanks @aharper273

karthikghan:
interestingly, the link doesn't work anymore. 

but now the original link works (  https://og-aws-slack.lexikon.io/ )

karthikghan:
[](https://og-aws.slack.com/join/shared_invite/enQtODM2NjY1NDQ2MTgxLWYwY2VjZDBiOGU1YTJjOWUwNTY3NjEyODA0NzY1N2MxNjhhZmYwZTU0NjNhMjNlNGVjODdlNTI4N2Y1YWIwNGE)

sneko:
Can confirmed what @karthikghan said, the link has expired :(

sneko:
Hey, I hope it will help someone, here is the right answer: https://github.com/open-guides/og-aws/pull/732#issuecomment-595199896

--> http://slackhatesthe.cloud/

neilthecellist:
[...]

PR #735, opened 2019-12-17
Update README.md

IsmaelMartinez:
Added extra link from https://github.com/open-guides/og-aws/issues/720#issuecomment-566734921

Once you are already in, either link works.


QuinnyPig:
Those links unfortunately expire and have to be recreated periodically. Slack hates us.

Therefore I've got a repointable domain for this: http://slackhatesthe.cloud 

Want to update your PR to use that domain and I'll merge it?

IsmaelMartinez:
Hi, reading the slack docs admin users can create a link that works for 2000 people. 

https://slack.com/intl/en-gb/help/articles/201330256#share-an-invitation-link

I rather prefer that in the README.md that a http link to who knows where. I know you are the owner of the domain now but you should not need to bother about it.

PR #748, opened 2020-05-18
Update certification list

juaoose:
Updated certification list, the big data certification is still listed as it will be available until july.


QuinnyPig:
Looks great--thanks!

PR #750, opened 2020-06-07
(fix) update zombie clause number

sw-yx:
The AWS TOS zombie clause number changed from 57.10 to 42.10. Updated and gave it a link for more context.


QuinnyPig:
So it did; thanks!

PR #751, opened 2020-06-08
Feat: added link to go to top 

itsjwala:
I found while reading some topic had to scroll straight up to see table of content, it got annoying after some time.

Added link to go to top after every `###` h3 heading, thought that would make it easier for navigating, can check the preview of the same in my [fork](https://github.com/itsjwala/og-aws)

Finally, Thanks for the awesome resource 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,6 @@ Legend
 General Information
 -------------------
 
-[Back to top :arrow_up:](#table-of-contents)
 ### When to Use AWS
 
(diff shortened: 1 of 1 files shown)


QuinnyPig:
I like this, but given its scope / scale, let's get other folks to weigh in! @maishsk, @jlevy, @ThanosBaskous: any thoughts?

QuinnyPig:
Thanks for the submission. Sorry it took so long to merge!

PR #759, opened 2020-11-02
Update ElastiCache URLs

pmarques:
There are in fact 2 pages for "select engine" although they look the same:
* https://docs.aws.amazon.com/AmazonElastiCache/latest/mem-ug/SelectEngine.html
* https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/SelectEngine.html

Just using one of them for simplicity.

**NOTE**: also changing url to HTTPS

### DIFF
--- a/README.md
+++ b/README.md
@@ -1395,7 +1395,7 @@ ElastiCache
 ### ElastiCache Tips
 
 - Choose the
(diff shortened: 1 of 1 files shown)


QuinnyPig:
Thanks for the contribution!

PR #767, opened 2020-12-20
Fixes slack URLs everywhere to correct url everywhere

acsrujan:
Hope @QuinnyPig never lets go of this domain. :)


QuinnyPig:
It's not going anywhere. Thanks!

PR #768, opened 2020-12-20
adds lambda gotcha on public subnets requiring NAT gateway

acsrujan:


QuinnyPig:
Thanks for the submission!

PR #774, opened 2021-02-16
Update README.md

sherihans123:
Added a resource page that will be appreciated by your audience.


darigovresearch (approved):
Looks good to us as it added a working link to a certification program, you may want to update the name of the pull request so it's more specific, take a look at the [contributing guidelines](https://github.com/open-guides/og-aws/blob/master/CONTRIBUTING.md) which may be helpful.

The checks have also expired so can't provide additional guidance on that

Do the maintainers need anything else for this to be merged?

QuinnyPig:
I'm conflicted on how to address these; it feels like it'd lead to a scenario where every certification / training provider could make a good case for inclusion. @jlevy, what're your thoughts?

darigovresearch:
@QuinnyPig you raise an interesting point, @jlevy what do you think?

PR #778, opened 2021-06-13
Add information about obtaining elastic IP blocks

kyl191:
AWS does carve out elastic IP blocks for (presumably) companies with larger spends
eg Boingo: https://search.arin.net/rdap/?query=54.241.191.47, Netflix: https://search.arin.net/rdap/?query=54.241.191.62

Not too sure if adding an example helps, but the ARIN data is a definitive record that it _does_ happen.


QuinnyPig:
Ooh, TIL. Thanks; great find!

PR #779, opened 2021-06-24
Fix misnomer of Identity and Access Management

noamsdahan:
Under S3, Identity and Access Management (IAM) is listed as Identity and Authentication Management.


QuinnyPig:
Good find--thanks!

PR #781, opened 2021-06-30
Switch broken link for Internet Archive copy

jciskey:
The current article link for S3 eventual consistency redirects to a Google page instead of the original article. This update switches out the link to the Internet Archive copy of the original article.


QuinnyPig:
Thanks; good catch!

PR #786, opened 2021-10-19
Update ECS links to hosted Kubernetes solutions

snay2:
Made the following updates:

* Google's offering is now called Google Kubernetes Engine and has a new URL
* AWS has a Kubernetes offering, EKS, already described in this document, so the Heptio reference implementation need not be listed with the other hosted solutions
* Azure, DigitalOcean, and OpenShift are already listed as alternatives in the EKS section and should be included here as well

Disclosure: I work for AWS.


QuinnyPig:
Well spotted; thanks!

PR #795, opened 2022-01-05
DynamoDB: Remove mention of ddb not supporting empty string

vladgolubev:
This is not relevant since 2020-05-18: https://aws.amazon.com/about-aws/whats-new/2020/05/amazon-dynamodb-now-supports-empty-values-for-non-key-string-and-binary-attributes-in-dynamodb-tables/


QuinnyPig:
Good find; thanks!

PR #798, opened 2022-02-17
Update README.md

mariuszmspace:


QuinnyPig:
Thanks, but this is clearly not aligned with our policy on self-promotion.

PR #803, opened 2022-05-27
Add tagging limitation on shared resources

nitrocode:
## what

- Add tagging limitation on shared resources

## why

- Document notable tagging limitation/gotcha

## references

- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sharingamis-explicit.html

> Tags – User-defined tags that you attach to a shared AMI are available only to your AWS account and not to the other accounts that the AMI is shared with.

- https://docs.aws.amazon.com/vpc/latest/tgw/tgw-transit-gateways.html

> (Optional) Under Tags, type a tag key and tag value pair for each tag. These tags are applied to the resource share but not to the transit gateway.

## notes

I cannot find a general reference for the limitation on the ability to share resources cross-account with the exception of sharing tags cross-account


darigovresearch (approved):
Thanks @nitrocode for making the pull request & providing an additional insight, think this is worth adding as the gotcha is value add and couldn't see any clashes with the [contributing guide](https://github.com/open-guides/og-aws/blob/master/CONTRIBUTING.md)

Are the maintainers happy to merge the pull request?

QuinnyPig:
Super interesting caveat. I really need to check this queue more often; merged!

PR #820, opened 2023-11-08
Updated RDS for SQL Server supported versions

w1wise:


danullom:
Can we also remove the RDS SQL Server "Gotchas and Limitations" that says storage cannot be expanded for existing databases?  That is no longer true.
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_PIOPS.StorageTypes.html#USER_PIOPS.ModifyingExisting

PR #827, opened 2024-08-16
Add Zuplo to API Gateway alternatives

AdrianMachado:
I think Zuplo is worth mentioning here as many people use it instead of AWS API gateway to avoid complexity and vendor lock-in


AdrianMachado:
@jlevy - any chance you can review?

